PR #1, opened 2017-07-21
Add query to ProviderServiceRequest model class when json is serialized

nicolasfagotti:
Hi,

From my experience using the library, if the query is not returned in the jsonSerialize result, then it won't be persisted in the pact file generated.

Please disregard the change if it makes no sense to you.

### DIFF
--- a/src/Mocks/MockHttpService/Models/ProviderServiceRequest.php
+++ b/src/Mocks/MockHttpService/Models/ProviderServiceRequest.php
@@ -146,6 +146,10 @@ class ProviderServiceRequest implements \JsonSerializable, \PhpPact\Mocks\MockHt
         $obj->method = $this->_method;
         $obj->path = $this->_path;
 
+        if ($this->_query) {
+            $obj->query = $this->_query;
+        }
+
         $header = $this->_headers;
         if (is_array($header)) {
             $header = (object)$header;


mattermack:
@nicolasfagotti sorry for the delay.  I wired up a CD process in AppVeyor (https://ci.appveyor.com/project/mattermack/pact-php)

bethesque:
Awesome! I was going to try and get a windows build running, but my Windows/python foo is very low. I'm still trying to work out how to get the mock service process shutting down on windows.

bethesque:
@mattermack could you add this to the pact-foundation project?

bethesque:
Oh, I beg your pardon. I'm working backwards through my notifications, so now I can see you've done it already.

mattermack:
@bethesque ,

If this help you solve your Windows problem, check out https://github.com/pact-foundation/pact-php/blob/master/example/test/bootstrap.php .  While it is in PHP, it does show how I was able to hack my way through it.

mattermack:
This should be in Composer v 1.0.5

bethesque:
@mattermack I can shut it down using force as you have, but it's a nasty thing to do to a process. It doesn't give it a chance to tidy up any resources. It feels dirty! I guess it doesn't matter as it's all in memory, but I'd really prefer to use a ctl+c equivalent if possible. I guess I can release it with the force kill for now and then improve it when I can. I've gotten this mostly working: https://gist.github.com/rdp/f51fb274d69c5c31b6be

Issue #7, opened 2017-10-06
Use arrays instead of stdClass


ruudk:
Any special reasons to use stdClass while doing json_decode? 

I found that there are typing issues because sometimes something is an array and sometimes its a stdClass. 

Can I create a PR to remove all stdClass usages and just use plain arrays instead?

mattermack:
@ruudk ,
We're definitely taking pull requests.  I just wrapped up a full testing on spec 1.1.   With the Pact foundatoin, I have AppVeyor setup to run the tests on Windows: https://ci.appveyor.com/project/mattermack/pact-php

Assuming the tests pass, pull requests are definitely welcome.  I appreciate you putting in these issues.  I'll have to work through them with the 1.1.0 version to see if they are still open.  If you could look at them again, let me know.

PR #9, opened 2017-10-06
Always return headers as array

ruudk:
See https://github.com/pact-foundation/pact-php/issues/7


mattermack:
@ruudk,

I just made a major commit as the older version was not passing all 1.1 specification tests.  Now, it should.  Can you test that the new implementation works?   I should have the new version in Packagist soon

ruudk:
Rebased!

mattermack:
Reverted.  See https://github.com/pact-foundation/pact-php/pull/14 on CI failures

PR #18, opened 2017-10-09
Use arrays instead of stdclass

ruudk:
Trying to fix #14 

But somehow 1 tests keeps on failing.

I have no clue why. I spend already 2 hours on it, maybe @mattermack has an idea? :) 

PS: Will rebase when I can get this to work.


mattermack:
If it makes you feel better, it took me a better part of Friday night to get the use cases to work.  I'll look at this as I can.

ruudk:
I think I can never let this test pass, because the differences between `json_decode` with the `assoc: true` flag.

It will create multidimansional arrays, and appearently https://github.com/pact-foundation/pact-php/blob/master/test/IntegrationTests/testcases-v1_1/response/body/deeply%20nested%20objects.json#L12 only cares for the inner arrays :o , kinda weird if you ask me.

mattermack:
@ruudk,

I'm working on Pact 2.0 with matchers.   Check out the branch if you are interested.

PR #30, opened 2018-01-16
Pact Refactor Part 1

nbrink91:
There are a few parts to this. I bumped the version up to 3 as there are some major breaking changes.

1. Added CS fixer via Composer. This will prevent the need to pull down a phar file.
2. Create a Ruby Standalone Wrapper.
3. Cleaned up the code a bit to follow PSR standards better.
4. Added a new example.


mattermack:
Deferring to Migrate Consumer to Ruby Backend #32

PR #33, opened 2018-01-18
Improved matchers to support array values. Squashed some bugs. Documentation improvements.

nbrink91:


example/tests/Provider/PactVerifyTest.php
```diff
@@ -0,0 +1,42 @@
...
+use Symfony\Component\Process\Process;
+
+class PactVerifyTest extends TestCase
+{
+    /** @var Process */
+    private $process;
+
+    protected function setUp()
+    {
+        $this->process = new Process('php -S localhost:8080 -t ../../src/Provider/public/');
```
mattermack: @nbrink91, this is to work around using a bootstrap?

nbrink91: Yeah, let me clean that up a bit. I did that and didn't wrap it up. I need to add an assertion as well.

src/PhpPact/Consumer/TestCase/MockServerTestCase.php
```diff
@@ -0,0 +1,36 @@
...
+    private $mockServer;
+
+    protected function setUp()
+    {
+        $this->config = new MockServerEnvConfig();
+        $this->config
+            ->setPactFileWriteMode('merge')
+            ->setPactDir('D:/Development/PACT/pact-php');
+        $this->mockServer = new MockServer($this->config);
+        $this->mockServer->start();
```
mattermack: I thought the start and stop was in the listener (i.e. reason to move away from the bootstrap)

src/PhpPact/Consumer/TestCase/MockServerTestCase.php
```diff
@@ -0,0 +1,36 @@
...
+    protected $config;
+    /** @var MockServer */
+    private $mockServer;
+
+    protected function setUp()
+    {
+        $this->config = new MockServerEnvConfig();
+        $this->config
+            ->setPactFileWriteMode('merge')
+            ->setPactDir('D:/Development/PACT/pact-php');
```
mattermack: Can this be a relative path?

src/PhpPact/Consumer/TestCase/MockServerTestCase.php
nbrink91: You are right, I added this when playing around with the merge ability of the ruby mock server. I deleted the file. Thanks!

src/PhpPact/Consumer/TestCase/MockServerTestCase.php
nbrink91: Yes, you can definitely use a relative path. This file should have never been committed. It was just something I was experimenting with.

src/PhpPact/PactBrokerConnector.php
```diff
@@ -59,6 +59,7 @@ public function publishFile($file, $version)
 
     /**
      * @deprecated This will be removed in the next major release. @see BrokerHttpService::publishJson for the replacement.
```
mattermack: Nice, @nbrink91

PR #35, opened 2018-01-23
Pact Ruby Standalone Verifier Implementation

nbrink91:
Also includes:

- TravisCI configuration for Linux testing.
- Updated ReadMe.md

Todos:

- Discuss integration tests with PACT broker
- Discuss turning on TravisCI
- Discuss dev releases


example/tests/Provider/PactVerifyTest.php
```diff
@@ -2,11 +2,11 @@
...
 namespace Provider;
 
-use GuzzleHttp\Client;
-use PhpPact\PactBrokerConnector;
-use PhpPact\PactUriOptions;
-use PhpPact\PactVerifier;
-use PhpPact\PactVerifierConfig;
+use GuzzleHttp\Psr7\Uri;
+use PhpPact\Broker\Service\BrokerHttpService;
+use PhpPact\Http\GuzzleClient;
```
mattermack: This project wraps the GuzzleClient?

mattermack:
I cannot review it b/c the PR is so big.  It looks like the tests pass

example/tests/Provider/PactVerifyTest.php
nbrink91: I made an interface to wrap the Http Client. That way we can swap from Guzzle to something else relatively easily in the future.

PR #53, opened 2018-03-14
Fix host setter in mock server config

samuelnogueira:


mattermack:
Thanks @samuelnogueira

PR #57, opened 2018-04-24
Fix deprecation warning in GuzzleHttp\Psr7\Uri by adding leading slashes

Lyhl:
I get deprecation warnings after uploading pacts to pact broker (after successull test execution).
Adding leading slashes like in this PR will solve this problem. 

see https://github.com/guzzle/psr7/blob/master/src/Uri.php#L693 
Remaining deprecation notices (1)

  1x: The path of a URI with an authority must start with a slash "/" or be empty. Automagically fixing the URI by adding a leading slash to the path is deprecated since version 1.4 and will throw an exception instead.
    1x in PactTestListener::endTestSuite from PhpPact\Consumer\Listener

In my project that leads on ci to a non-zero exit code of my script


mattermack:
Thanks, @Lyhl

PR #63, opened 2018-04-27
test also lowest available dependencies

mente:
To verify that pact-php works also with minimal declared dependencies (which appeared it's not).

Original intention was to lower `symfony/*` dependencies to be able to use in symfony 2.x projects. But after that I've realised that dependencies should be not lowered but bumped up :-/.

If I can contribute PR for support of symfony 2.x (yay, we have some legacy here) will it have chances to be accepted?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,20 @@ os:
 
 language: php
 
-php:
-  - 7.0
-  - 7.1
-  - 7.2
+matrix:
+    include:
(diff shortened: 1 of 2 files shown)


nbrink91:
@mente, thanks for adding some extra testing validation. Feel free to pursue Symfony 2 compatibility. Assuming that the symfony/process repo in version 2 isn't dramatically different it may not be too difficult.

PR #68, opened 2018-05-02
use phar distribution of php-cs-fixer

mente:
Another step towards symfony 2.x support. `php-cs-fixer` has dependencies on symfony 3.x . By using phar instead dependencies are not being leaked to the overall library. This PR includes [tm/tooly-composer-script](https://github.com/tommy-muehle/tooly-composer-script) which actually just installs the same phar in the end of the composer command.

I've also updated composer scripts to use local phpunit and php-cs-fixer instead of global one

PS It looks like `composer fix` haven't been run for a while, but I leave it up to maintainers to define/revisit CS rules and CI pipeline for it

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,9 +34,9 @@
     "require-dev": {
         "roave/security-advisories": "dev-master",
         "phpunit/phpunit": "^6.4",
-        "friendsofphp/php-cs-fixer": "^2.10",
         "mockery/mockery": "^1.0",
-        "slim/slim": "^3.9"
+        "slim/slim": "^3.9",
+        "tm/tooly-composer-script": "^1.2"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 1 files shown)


mente:
appveyor is failing but on the other side it duplicates travis functionality, it's `composer test` in the end. What's point of appveyor then? Maybe it's worth removing?

mente:
Or should I invest into fixing the problem?

mente:
My fault, appveyor is for windows and `vendor/bin` is automatically included in `$PATH`. Reverted and rebased

nbrink91:
I'm going to go ahead and hold out on merging this PR in until we get Symfony 2 working as expected in a feature branch. I added a PHPCS lint to the Travis CI build that will now fail the build if the syntax errors haven't been fixed.

Issue #76, opened 2018-05-31
PhpPact\Standalone\ProviderVerifier\Verifier verification issue


arkonchik:
In 3.1 version I can verify provider pact in next ways:

1. Verification problem
- verify latest tag for provider-consumer
```
public function verify(string $consumerName, string $tag = null): self
    {
        // Add tag if set.
        if ($tag !== null) {
            $uri = $this->config->getBrokerUri()
                ->withPath("pacts/provider/{$this->config->getProviderName()}/consumer/{$consumerName}/latest/{$tag}");
        } else {
            $uri = $this->config->getBrokerUri()
                ->withPath("pacts/provider/{$this->config->getProviderName()}/consumer/{$consumerName}/latest");
        }
```
 
- verify all consumers for provider 
```
`public function verifyAll()
    {
        $arguments = $this->getBrokerHttpClient()->getAllConsumerUrls($this->config->getProviderName(), $this->config->getProviderVersion());

```
- verify all consumers for tag
```
public function verifyAllForTag(string $tag)
    {
        $arguments = $this->getBrokerHttpClient()->getAllConsumerUrlsForTag($this->config->getProviderName(), $tag);
```

But how can I verify pact for certain consumer version?
This doc https://github.com/pact-foundation/pact_broker/wiki/Pacticipant-version-numbers told that consumer version number can be git sha, but for now I can't verify that provider supports pact for consumer with certain version, through this PhpPact\Standalone\ProviderVerifier\Verifier - class 
Maybe need another one method like?
`public function verifyVersion(string $consumerName, string $version = null): `

2. Property set up problem
Can we move this from const to var and set up from constructor at Verifier?
```
const PROCESS_TIMEOUT      = 60;
const PROCESS_IDLE_TIMEOUT = 10;
```

arkonchik:
Maybe it will better if I do pull request?

mattermack:
Playing with Pact Broker, I think one cannot use both a tag and a consumer version.   Is that right @bethesque ?  

Assuming that is true, a simple code change is sufficient. 

```
public function verify(string $consumerName, string $tag = null, string $consumerVersion = null): self
{
        $path = "pacts/provider/{$this->config->getProviderName()}}/consumer/{$consumerName}/";
        if ($tag) {
            $path .= "latest/{$tag}/";
        }
        else if ($consumerVersion) {
            $path .= "version/{$consumerVersion}/";
        } else {
            $path .= "latest/";
        }

[...]

PR #78, opened 2018-06-17
Added configuration for timeout handling.

siad007:
Related to #76

### DIFF
--- a/src/PhpPact/Consumer/Listener/PactTestListener.php
+++ b/src/PhpPact/Consumer/Listener/PactTestListener.php
@@ -18,7 +18,7 @@ use PHPUnit\Framework\TestSuite;
 
 /**
  * PACT listener that can be used with environment variables and easily attached to PHPUnit configuration.
- * Class PactTestListener
+ * Class PactTestListener.
  */
 class PactTestListener implements TestListener
 {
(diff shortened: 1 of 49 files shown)


src/PhpPact/Consumer/Listener/PactTestListener.php
```diff
@@ -58,20 +58,20 @@ public function __construct(array $testSuiteNames)
      *
      * @throws \Exception
      */
-    public function startTestSuite(TestSuite $suite)
+    public function startTestSuite(TestSuite $suite): void
```
samuelnogueira: This library should be compatible with PHP 7.0, but void return was introduced only in PHP 7.1 (see http://php.net/manual/en/migration71.new-features.php).

src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfig.php
```diff
@@ -42,16 +42,20 @@ class VerifierConfig implements VerifierConfigInterface
     /** @var string */
     private $format;
 
+    private $processTimeout = 60;
```
samuelnogueira: Missing var type declaration `/** @var int */`

PR #86, opened 2018-07-31
Feat/message pact

mattermack:
@nbrink91 please review if you can


mattermack:
This does move the Symphony Process dependency from 2.8 to 3.3.  I would like to move completely off of Symphony to Amp.   Amp is needed for the message functionality, solving an async case that Symphony, React, and others could not solve on Windows.

mattermack:
@nbrink91 , thanks.  I made the changes you suggested.

PR #89, opened 2018-09-03
Loosen symfony/console version bounds

samuelnogueira:
Since symfony/console uses [semver](https://semver.org/), it's best to use the composer `^` operator instead of strict upper bounds on version.
This allows this library to play nice with other libraries requiring symfony/console versions 4.1 and 4.2

See:
https://getcomposer.org/doc/faqs/why-are-unbound-version-constraints-a-bad-idea.md


cfmack:
Thanks @samuelnogueira

PR #90, opened 2018-09-07
make health check timeout of mock server configurable

frastel:
In my local setup I am not able to start the MockServer within 10 retries/seconds. It takes about 12-14 seconds.
After I patch the allowed timeout in the health check directly everything works as expected.

For better stability I want to define this timeout on my own. Other ppl out there probably too.
Therefore I introduced the env var `PACT_MOCK_SERVER_HEALTH_CHECK_TIMEOUT`. When this env var is not set the timeout the used timeout of 10 retries/seconds is used.

### DIFF
--- a/example/phpunit.consumer.xml
+++ b/example/phpunit.consumer.xml
@@ -24,6 +24,7 @@
         <env name="PACT_CONSUMER_TAG" value="master"/>
         <env name="PACT_PROVIDER_NAME" value="someProvider"/>
         <env name="PACT_OUTPUT_DIR" value=".\example\output"/>
+        <env name="PACT_MOCK_SERVER_HEALTH_CHECK_TIMEOUT" value="10"/>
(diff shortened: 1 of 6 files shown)


mattermack:
Thanks @frastel

bethesque:
That's a ridiculously long time. I've raised an issue to see if we can work out why it takes that long.

bethesque:
Hm. I may not be able to reproduce it. It takes 3 seconds (which is still too long, but not as bad) on my machine.

```
$ time pact/bin/pact-mock-service start -p 1234
[2018-09-08 11:32:11] INFO  WEBrick 1.3.1
[2018-09-08 11:32:11] INFO  ruby 2.2.2 (2015-04-13) [x86_64-darwin13]
[2018-09-08 11:32:11] INFO  WEBrick::HTTPServer#start: pid=34120 port=1234

real	0m3.254s
user	0m0.421s
sys	0m0.193s
```

PR #96, opened 2018-10-10
make latest Symfony 4 components installable

frastel:
Currently it is not possible to install this library within a new Symfony 4 application (>= 4.1!). This has huge impact on the acceptance for this library in our company.
Besides I see no reason why the Symfony 4 components should be pinned to only < 4.0 and therefore I loose the requirements [once more](https://github.com/pact-foundation/pact-php/commit/527184daab0404a9380e2a277c5b08ae83795be4#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780).
All upcoming Symfony 4.* versions are installable now.

I run the integration tests in the `examples` directory to check if something broke, but everything looks good.

In addition I moved the pact-broker to port `:8080` as port `:80` is often already in use on a host machine.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,8 +20,8 @@
     "require": {
         "php": "^7.1",
         "ext-openssl": "*",
-        "symfony/process": ">=3.3 <4.1",
-        "symfony/filesystem": ">=2.8 <4.1",
+        "symfony/process": "^3.3 || ^4.0",
+        "symfony/filesystem": "^2.8 || ^3.0 || ^4.0",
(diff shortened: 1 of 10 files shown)


mattermack:
This is similar to  https://github.com/pact-foundation/pact-php/commit/527184daab0404a9380e2a277c5b08ae83795be4#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780

frastel:
Thanks @mattermack. I can confirm that pact-php is installable within a fresh Symfony 4.1 application now!

Issue #103, opened 2018-11-13
Allow disabling of ssl-verification for self-signed cert at broker


eiriarte-mendez:
When using a broker with a self-signed cert, pact is not able to upload the contract. 

```
In CurlFactory.php line 186:                                                                                                                                          
  cURL error 60: SSL certificate problem: unable to get local issuer certificate (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)  
```

PR #104, opened 2018-11-13
configurable ssl configuration for broker

eiriarte-mendez:
Defines a env-variable (`PACT_BROKER_SSL_VERIFY`) that allows to disable SSL-verification .
It solves #103


cfmack:
Thanks, @eiriarte-mendez

Issue #107, opened 2018-11-18
Pact files not written any more


DerManoMann:
Hi there,

using any commits after https://github.com/pact-foundation/pact-php/commit/4004e443186737b9d6ae4cc61f42228c20f70331 break my test code in that pact files are no longer written to disk. Not sure how that would be affected but just switching back to the above commit makes things work again.
Any pointers as to how to debug this would be appreciated.

cirnatdan:
Might be my PR that caused this (https://github.com/pact-foundation/pact-php/pull/98)
Will check this and let you know

cfmack:
Wow, we have so many tests and missed this one :).   I am going to be traveling and won't be able to look into this for a few days.

cfmack:
I think I can replicate this, @DerManoMann .  

1. The examples were written on Windows
1. The output pacts that the consumer tests generate were checked in
1. The phpunit.consumer.xml and phpunit.all.xml explicitly had Windows pathing: "PACT_OUTPUT_DIR" value=".\example\output\\"
1. The mock server is an instance of the Ruby server

Even when replacing the \ in the XML, .\example\output was called to the Ruby server.   When switching to Amp, somehow I assume, the directory pathing was not correct.   When running on Linux (which I assume you are doing), Ruby strips out the \ and the file output becomes .exampleoutput.   

I assume you have something similar as you noted the "defaults changed", which they seemed to by stripping out the extra \ .  

To be clear, this is *not* a bug of the Ruby server.

cfmack:
@DerManoMann , I modified the pathing code and examples on 

explicitly changed pathing when calling ruby server #108

If this works for you, I'll make another tag.  Right now, it is hanging out on Master

DerManoMann:
Hmm, tested this just now with
````
pact-foundation/pact-php           dev-master 3c234e7 Enables consumer driven contract testing, following the PACT foundation principles.
````
and still not seeing any files.
The CL looks like this and the path names are all correct. I also debugged and the change in `setPactDir()` doesn't seem to be making a difference anyway on linux.
````
server.DEBUG: Process command: /.../vendor/pact-foundation/pact-php/src/PhpPact/Standalone/Installer/../../../../pact/bin/pact-mock-service service --consumer=php-client --provider=api-service --pact-dir=/../pacts --pact-file-write-mode=merge --host=localhost --port=7200 --pact-specification-version=2 [] []
````

[...]

PR #112, opened 2018-12-01
Issue #107 - correctly close the Ruby process

cfmack:
Change kill 09 to kill.   
Add wrapper around getPactJson to explicitly write the file - writePact()


cfmack:
Code fixes for #107

Testing still in progress as struggling with php-unit

PR #113, opened 2018-12-12
Fix Broker HTTP Authentication

cirnatdan:
While trying to integrate Pact consumer tests into my existing CI, I couldn't get HTTP auth working with the Pact Broker.

Seems like a typo and a quick fix to me


cfmack:
Thank you

PR #123, opened 2019-05-28
Verifier output

Thomblin:
This pull request ensures that all output is captured, if `VerifierProcess:run()` executes any command that fails (exit code >0)

There are two problems currently:

1. The output is not passed to the logger, if `$processRunner->runBlocking()` throws an exception
2. `ProcessRunner:run()` does not capture the complete output, if stdout and stderr are used alternating (see verifier.sh script in UnitTest)

### DIFF
--- a/example/src/Provider/public/index.php
+++ b/example/src/Provider/public/index.php
@@ -10,13 +10,17 @@ $app = new \Slim\App();
 $app->get('/hello/{name}', function (Request $request, Response $response) {
     $name = $request->getAttribute('name');
 
-    return $response->withJson(['message' => "Hello, {$name}"]);
+    return $response
+            ->withJson(['message' => "Hello, {$name}"])
+            ->withHeader('Content-Type', 'application/json');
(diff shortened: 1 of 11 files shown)


cfmack:
Getting this to run right on both Linux and Windows has proven difficult for me.  I welcome the assistance.

Thomblin:
You are right. I was not able to get stable results so far. I'll have to take a closer look at Amp Process. I currently try using the latest release. Any help is welcome

Thomblin:
@cfmack I guess it is ready for review now. However the ProcessRunner could need some refactoring regarding the usage of logger an run / runBlocking in my opinion.

cfmack:
Good work.   The runBlocking and run functions were added as part of the move from Symphony to Amp.   The need for Amp was required for Windows multiple process support, which Symphony did not offer.   Thanks for the PR

cfmack:
Tagged as 5.0.4

PR #125, opened 2019-05-30
Loose logging

Thomblin:
As a developer I want to have full control of all logs that are written. Thus I added a possibility to inject Logger instances in several classes


mattermack:
I still got an odd error, even when trying to merge this in after #126 https://ci.appveyor.com/project/mattermack/pact-php/builds/25020714

mattermack:
You're right.  It is unreliable.  After running again, I got it to pass.   Based on this, I'll merge these changes into master but not tag a branch.   I'll reopen that ticket you opened

Issue #124, opened 2019-05-30
Unstable test


Thomblin:
Even if my pull request was green, the merge into master left one test failing.

> Test 'PhpPact\Standalone\Runner\ProcessRunnerTest::testBlockingProcess' started
> [2019-05-30 02:54:05] server.DEBUG: Process command: dir  [] []
> [2019-05-30 02:54:05] server.DEBUG: Exit code: 0 [] []
> [2019-05-30 02:54:05] server.DEBUG: Process command: dir failed.xx [] []
> [2019-05-30 02:54:05] server.DEBUG: Exit code: 2 [] []
> Test 'PhpPact\Standalone\Runner\ProcessRunnerTest::testBlockingProcess' ended
> 
> There was 1 failure:
> 
> 1) PhpPact\Standalone\Runner\ProcessRunnerTest::testBlockingProcess
> Expect 'failed.xx' to be in the stderr
> Failed asserting that false is true.
> C:\projects\pact-php\tests\PhpPact\Standalone\Runner\ProcessRunnerTest.php:47

Any ideas how to fix it?

mattermack:
Merged based on your refactor

mattermack:
Reopenning as Windows continues to appear unstable:
Reason 1: https://ci.appveyor.com/project/mattermack/pact-php/builds/25020714
Reason 2: https://ci.appveyor.com/project/mattermack/pact-php/builds/25020861
Passing (w/o changes): https://ci.appveyor.com/project/mattermack/pact-php/builds/25020935

Thomblin:
That is strange. I'll take a look at it

Thomblin:
I created a pull request

PR #126, opened 2019-06-02
Fix issue #124 unstable windows processes

Thomblin:
This fixes windows processes and allows them to run in background
I had to upgrade the Amp packages


mattermack:
Thank you!

Issue #131, opened 2019-06-19
Broker triggers Provider CI even if the pact is not changed


enesanbar:
I create a webhook to trigger Provider CI job in the Pact Broker as follows:

```json
{
  "Consumer": {
    "name": "Foo"
  },
  "provider": {
    "name": "Bar"
  },
  "events": [{
    "name": "contract_content_changed"
  }],
  "request": {
    "method": "POST",
    "url": "http://master.ci.my.domain:8085/rest/api/latest/queue/SOME-PROJECT?os_authType=basic",
    "username": "username",
    "password": "password",
    "headers": {
      "Accept": "application/json"
    }
  }
}
```

As seen, I only use "contract_content_changed" event in the Webhook. When consumer CI job sends the _unchanged_ pact file to the Broker, the Provider CI job gets triggered no matter what. 

After I did some digging, I found out that PactTestListener tags the consumer after publishing the pact to the broker, by which time the event is already triggered on the Broker without tags applied to the consumer.

PR #132, opened 2019-06-19
Tag the Consumer before publishing the Pact to the Broker

enesanbar:
Closes: #131

### DIFF
--- a/src/PhpPact/Consumer/Listener/PactTestListener.php
+++ b/src/PhpPact/Consumer/Listener/PactTestListener.php
@@ -118,8 +118,8 @@ class PactTestListener implements TestListener
                 $client = new GuzzleClient($clientConfig);
 
                 $brokerHttpService = new BrokerHttpClient($client, new Uri($pactBrokerUri));
-                $brokerHttpService->publishJson($json, $consumerVersion);
                 $brokerHttpService->tag($this->mockServerConfig->getConsumer(), $consumerVersion, $tag);
+                $brokerHttpService->publishJson($json, $consumerVersion);
                 print 'Pact file has been uploaded to the Broker successfully.';
             }
         }


src/PhpPact/Consumer/Listener/PactTestListener.php
```diff
@@ -118,8 +118,8 @@ public function endTestSuite(TestSuite $suite): void
                 $client = new GuzzleClient($clientConfig);
 
                 $brokerHttpService = new BrokerHttpClient($client, new Uri($pactBrokerUri));
-                $brokerHttpService->publishJson($json, $consumerVersion);
                 $brokerHttpService->tag($this->mockServerConfig->getConsumer(), $consumerVersion, $tag);
+                $brokerHttpService->publishJson($json, $consumerVersion);
```
cfmack: What if this is a new consumer?   Does tag create the Consumer?   Does the broker create the consumer in the tag function before there is an initial contract?

enesanbar: Yes, broker takes care of that. "tag" function just makes a PUT request to this path:
"/pacticipants/{$consumer}/versions/{$version}/tags/{$tag}", which in turn creates the consumer if it doesn't exist.

```sh
postgres=# select * from pacticipants;
 id | name | repository_url | created_at | updated_at
----+------+----------------+------------+------------
(0 rows)
```

```sh
$ curl -X PUT -H "Content-Type: application/json" http://pact-broker/pacticipants/non-existing-consumer/versions/1.0.0/tags/master/ -i
HTTP/1.1 201 Created
Content-Type: application/hal+json;charset=utf-8
Content-Length: 510
Connection: keep-alive
Status: 201 Created
Location: http://pact-broker/pacticipants/non-existing-consumer/versions/1.0.0/tags/master
Date: Thu, 20 Jun 2019 05:47:35 GMT
X-Content-Type-Options: nosniff
Server: Webmachine-Ruby/1.5.0 Rack/1.3
X-Pact-Broker-Version: 2.31.0
X-Powered-By: Phusion Passenger 6.0.2

{"name":"master","updatedAt":"2019-06-20T05:47:35+00:00","createdAt":"2019-06-20T05:47:35+00:00","_links":{"self":{"title":"Tag","name":"master","href":"http://pact-broker/pacticipants/non-existing-consumer/versions/1.0.0/tags/master"},"version":{"title":"Version","name":"1.0.0","href":"http://pact-broker/pacticipants/non-existing-consumer/versions/1.0.0"},"pacticipant":{"title":"Pacticipant","name":"non-existing-consumer","href":"http://pact-broker/pacticipants/non-existing-consumer"}}}
```

```sh
postgres=# select * from pacticipants;
 id |         name          | repository_url |         created_at         |         updated_at
----+-----------------------+----------------+----------------------------+----------------------------
  3 | non-existing-consumer |                | 2019-06-20 05:47:35.663317 | 2019-06-20 05:47:35.663334
(1 row)
```

bethesque:
The tag was always designed to come first (autocreating the consumer), but for an unintended reason, it used to handle it ok if the tag came second. I made a change recently that means that the tag has to come first, and so it's brought to light all the libraries that had the ordering the wrong way around. I didn't realise there were so many places not using the "official" pact-broker-client.

If I was designing the API 6 years ago knowing what I know now, it wouldn't have this ordering dependency! You live and learn.

cfmack:
Thanks y'all

Issue #135, opened 2019-09-16
Contact Verification not compatible to Monolog ~2.0


frastel:
The logging mechanism in VerifierProcess is not compatible to Monolog ~2.0.

My suggestion: Always develop against LoggerInterface and not against concrete classes.

PR is upcoming!

PR #136, opened 2019-09-16
fix logging during verification

frastel:
Fixes https://github.com/pact-foundation/pact-php/issues/135.

The logger methods used during the verification are not included in the standardized LoggerInterface. These methods (addInfo, addError) are contained in Monolith v1 but not in Monolog v2. Thus the verification will break in any fresh installed Symfony 4.3 project.

I changed the usage of the logger to use the interface. It is always a bad practice to rely on concrete classes.

I also introduced ProcessRunnerFactory for better unit testing possibilities.


cfmack:
This is awesome.  Great work!

PR #137, opened 2019-11-14
Adding possibility to add server log path config from env/phpunit xml

Nox77:
To allow to set server log path from phpunit.xml


cfmack:
Thank you!

PR #138, opened 2019-11-28
Add bearer token authentication

basis42:
This PR adds bearer token authentication option in BrokerHttpClient(as used in service https://pactflow.io for example).


basis42:
CS ... will fix this and open a new pull request.

PR #141, opened 2019-12-04
Download contract from pact mock server hook

dabrowiecki:
We are not able to use an included pact-mock server in our CI pipeline (our php images are based on php alpine) due to the following issues:
- https://github.com/pact-foundation/pact-mock-service-npm/issues/18
- https://github.com/pact-foundation/pact-js/issues/305

In order to overcome this problem, we use a separate standalone pact-mock server. Unfortunately, the existing Consumer\Listener\PactTestListener is not designed to fetch the contract from the pact-mock server. I created this PHPUnit hook to fetch and store contract in the configured location, so that the file can be used as an artifact.

### DIFF
--- a/src/PhpPact/Consumer/Hook/ContractDownloader.php
+++ b/src/PhpPact/Consumer/Hook/ContractDownloader.php
@@ -2,14 +2,13 @@
 
 namespace PhpPact\Consumer\Hook;
 
-use PHPUnit\Runner\AfterLastTestHook;
+use Exception;
+use PhpPact\Http\GuzzleClient;
 use PhpPact\Standalone\MockService\MockServerEnvConfig;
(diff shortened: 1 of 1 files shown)


cfmack:
The CI builds were not passing.   It looks like the lint checks failed.

https://travis-ci.org/pact-foundation/pact-php/builds/620689564?utm_source=github_status&utm_medium=notification
https://ci.appveyor.com/project/mattermack/pact-php/builds/29312248

Run the following and resubmit the PR.
`php .\vendor\tm\tooly-composer-script\bin\php-cs-fixer.phar fix --config .php_c`

cfmack:
Thank you for the PR, btw!

dabrowiecki:
@cfmack Changes applied and the pipeline is passing.

PR #143, opened 2020-01-14
Allow setting provider version tag

cwdt:
This PR gives the allows setting the providers' version tag while verifying.

### DIFF
--- a/src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfig.php
+++ b/src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfig.php
@@ -21,6 +21,9 @@ class VerifierConfig implements VerifierConfigInterface
     /** @var string */
     private $providerVersion;
 
+    /** @var null|string */
+    private $providerVersionTag;
+
     /** @var bool */
     private $publishResults = false;
 
@@ -116,6 +119,24 @@ class VerifierConfig implements VerifierConfigInterface
     public function setProviderVersion(string $providerVersion): VerifierConfigInterface
     {
         $this->providerVersion = $providerVersion;
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getProviderVersionTag()
+    {
+        return $this->providerVersionTag;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setProviderVersionTag(string $providerVersionTag): VerifierConfigInterface
+    {
+        $this->providerVersionTag = $providerVersionTag;
 
         return $this;
     }
--- a/src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfigInterface.php
+++ b/src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfigInterface.php
@@ -59,6 +59,18 @@ interface VerifierConfigInterface
     public function setProviderVersion(string $providerAppVersion): self;
 
     /**
+     * @return null|string providers version tag
+     */
+    public function getProviderVersionTag();
+
+    /**
+     * @param string $providerVersionTag providers version tag
+     *
+     * @return VerifierConfigInterface
+     */
+    public function setProviderVersionTag(string $providerVersionTag): self;
+
+    /**
      * @return bool are results going to be published
      */
     public function isPublishResults(): bool;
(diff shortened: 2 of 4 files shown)


cfmack:
We have to work through https://github.com/pact-foundation/pact-ruby-standalone/issues/43.   You can help by creating a fork of https://github.com/pact-foundation/pact-ruby-standalone-e2e-example with the appropriate example

cfmack:
Well, I got past the Consumer side as the below produces what appears to be the correct pact

```
 pact-message update --consumer=test_consumer --provider=test_provider --pact-dir=/home/cmack/pact-php/example/tests/MessageConsumer/../../output/ '{"description":"an alligator named Mary exists","providerStates":[{"name":"a message", "params" : ["foo", "bar"]}],"metadata":{"queue":"wind cries","routing_key":"wind cries"},"contents":{"text":"Hello Mary"}}'
```

Provider is still failing with validation but could still be on PHP

cfmack:
The latest commit should functionally pass.  However, I think there are use cases on Message support that do not resonate at this point

- What happens with multiple call backs / provider states for a single pact?  Does that change the workflow?
- What do the providerState params actually do?  Right now it is essentially ignored, which seems okish, kinda

cfmack:
https://travis-ci.org/pact-foundation/pact-php/builds/637788483?utm_medium=notification&utm_source=github_status passed Travis. 

https://ci.appveyor.com/project/mattermack/pact-php/builds/30149627 passed on AppVeyor

cwdt:
> * What happens with multiple call backs / provider states for a single pact?  Does that change the workflow?

If I understood correctly, for the first bullet-point the idea is to not key-by provider state but by description right?

And for the second bullet-point, I think it should be passed into the callback. I tried to get this working in your branch but I get some weird behavior. 

I would expect the variable `$providerState` (at https://github.com/pact-foundation/pact-php/compare/uprev-pact#diff-5e78f32bb13e5bc5774c5fdd5e88ccdeR185) to have a `params` property with `foo` in there. Except this isn't the case, it is just empty.

cwdt:
I took a shot on keying by description based on your work. You can see the diff here: https://github.com/pact-foundation/pact-php/compare/uprev-pact...cwdt:uprev-pact

(Travis build succeeds: https://travis-ci.org/cwdt/pact-php/builds/637987393)

Although I still have the issue with the params being empty, I've set up a branch to demonstrate that: https://github.com/cwdt/pact-php/compare/uprev-pact...cwdt:uprev-pact-params-issue

(Travis build fails: https://travis-ci.org/cwdt/pact-php/builds/637990027)

bethesque:
Btw, something to keep in mind is that the message "building" code that is keyed by description should also be scoped by consumer, so that if you have two different consumers using the same description but different example data, they don't clash.

bethesque:
> What happens with multiple call backs / provider states for a single pact? Does that change the workflow?

At the moment, there will be one callback for each provider state. I'd like to add the option to just send them all at once, because I know there are people who need to access data from between states. But that could come later, and be a configurable option.

> What do the providerState params actually do? Right now it is essentially ignored, which seems okish, kinda

They should be passed in to the code that builds the provider state. They're used for things like specifying dynamic data so that you can make data that matches the expectations. To be honest, I've never used them personally, because all the matches on the provider side should really be type based anyway, so the actual values shouldn't matter too much. But they're in there now!

cfmack:
@cwdt ,

I merged the update to the Ruby library into master in PR #145.  All the tests were passing and some latent bugs were exposed.  I put in issue  #146 to account for these bugs which are needed before I tag master with a breaking change/uprev.

You can go ahead and put in a PR to add the feature you were looking for as the Ruby standalone version should support your need.

cwdt:
Thanks! I updated the PR by rebasing it to the new master. It allows setting the `--provider-version-tag=` option now.

bethesque:
So, just to make life difficult, you can set multiple provider version tags. (Same for consumer version tags when you publish a pact).

cfmack:
Thank you @cwdt

bethesque:
I noticed this was merged, but I think there's still an outstanding issue. Can multiple provider version tags be supplied?

bethesque:
Just seen there's a separate issue been raised for it - ignore above comment! Thanks.

PR #148, opened 2020-03-03
Fix custom provider headers enclosing

aymanalzarrad:
Currently custom provider headers are not being enclosed in quotes when passed to `pact-provider-verifier`.

For example, setting: `--custom-provider-header=Authorization: Bearer MyToken`, causes the header to be cut at the first space resulting in the `Authorization` header to have an empty value when parsed by `pact-provider-verifier`.


mattermack:
Thank you!

PR #151, opened 2020-05-28
json file is actually located in the pacts folder

nickdietel:


cfmack:
Thanks @nickdietel .  I wrapped this into PR #158 .  I had some other clean up to do.

PR #154, opened 2020-06-23
feat: update standalone to 1.86.0

bethesque:
Please expose the --log-dir and --log-level options to the native API.


cfmack:
As always, thanks @bethesque

PR #157, opened 2020-07-01
feat: include brokerToken param verifier config

lewisPrescott707:
Please include --brokerToken when verifying pacts

@cfmack @mattermack ?? You seem to be on the case with recent PR's


cfmack:
Awesome.  Thanks @lewisPrescott707

cfmack:
I have a few other open changes and I'll work to tag it once I get through those.

cfmack:
Tagged to 5.0.7.

Issue #150, opened 2020-05-08
when classmap authoritative true on composer, deprecation warning


lucagentile:
Hello,

when composer autoloader has authoritative class maps set on true, `league/uri-schemes` a dependency of `"amphp/http-server": "^1.0.1"` gives some warnings.

`Deprecation Notice: Class League\Uri\Schemes\AbstractUri located in ./vendor/league/uri-schemes/src/Schemes/deprecated.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class League\Uri\Schemes\Data located in ./vendor/league/uri-schemes/src/Schemes/deprecated.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class League\Uri\Schemes\File located in ./vendor/league/uri-schemes/src/Schemes/deprecated.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class League\Uri\Schemes\Ftp located in ./vendor/league/uri-schemes/src/Schemes/deprecated.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class League\Uri\Schemes\Http located in ./vendor/league/uri-schemes/src/Schemes/deprecated.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class League\Uri\Schemes\Uri located in ./vendor/league/uri-schemes/src/Schemes/deprecated.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
[...]

PR #159, opened 2020-07-04
when classmap authoritative true on composer, deprecation warning #150

cfmack:
Updated amp and modified failed message provider tests


cfmack:
Lots of dependencies were updated, especially to support PHP 7.2 low dependencies.

PR #165, opened 2020-09-15
feat: Allow broker token to be set

jbarton123:
Some brokers require the use of a broker token. This adds the ability for a user to set their broker token for verification.

### DIFF
--- a/src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfig.php
+++ b/src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfig.php
@@ -31,6 +31,9 @@ class VerifierConfig implements VerifierConfigInterface
     private $brokerUri;
 
     /** @var null|string */
+    private $brokerToken;
+
+    /** @var null|string */
     private $brokerUsername;
 
     /** @var null|string */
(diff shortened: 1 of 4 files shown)


src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfig.php
```diff
@@ -183,6 +186,24 @@ public function setBrokerUri(UriInterface $brokerUri): VerifierConfigInterface
         return $this;
     }
 
+    /**
+     * {@inheritdoc}}
```
mefellows, line 190: extra `}` ?

src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfig.php
```diff
@@ -183,6 +186,24 @@ public function setBrokerUri(UriInterface $brokerUri): VerifierConfigInterface
...
+     */
+    public function getBrokerToken(): ?string
+    {
+        return $this->brokerToken;
+    }
+
+    /**
+     * {@inheritdoc }
```
mefellows, line 198: space?

mefellows:
Thanks for this @jbarton123 - i'll let @mattermack complete his review, but at first glance it looks like it should work nicely.

cfmack:
Thank you @jbarton123 !  Appreciate it.

PR #166, opened 2020-09-16
fix: set the broker token on the broker http client

jameshalsall:


cfmack:
@jameshalsall , can you run the following to pass the lint checks and resubmit:
php .\vendor\tm\tooly-composer-script\bin\php-cs-fixer.phar fix --config .php_cs

cfmack:
Thanks @jameshalsall .  Apologies for the delay.  Chalk it up to just another reason 2020 is a brutal year.

PR #167, opened 2020-09-29
Fixed fatal error.

siad007:
Fixed missing code argument to prevent a fatal error.
See https://www.php.net/manual/de/exception.construct.php


cfmack:
Good catch @siad007

PR #169, opened 2020-10-13
Added support for multiple provider and consumer version tags.

### LINKED ISSUE #147: Allow support for multiple provider and consumer version tags
Allow setting provider version tag #143 enables you to tag provider versions with a tag.  Consumer side functionality should have existed (I think) already.  

Apparently multiple tags can be set at the same time.  This is to implement that functionality in PACT-PHP

siad007:
Fixes and closes #147

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
     "require": {
         "php": "^7.2",
         "ext-openssl": "*",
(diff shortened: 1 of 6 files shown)


cfmack:
Amazing.   Thank you for jumping on this!

PR #172, opened 2020-10-15
Escaped shell args.

### LINKED ISSUE #171: Add support for consumer/provider names with a space.
As in the given snipped, it is not easy to pass a valid consumer/provider name with a space.
For example `--consumer=Foo Bar` is not possible:

https://github.com/pact-foundation/pact-php/blob/3a29da6a8af6d9ffc11e808fa5850a8e7d23aefa/src/PhpPact/Standalone/MockService/MockServer.php#L108-L109

Before encoding it by hand or with `rawurlencode`, to convert the spaces to `%20`, we should consider to `rawurlencode` or maybe better `escapeshellarg` ❓ it right at that place.

What do you think?

siad007:
Fixes #171

### DIFF
--- a/src/PhpPact/Standalone/MockService/MockServer.php
+++ b/src/PhpPact/Standalone/MockService/MockServer.php
@@ -104,10 +104,14 @@ class MockServer
     {
         $results = [];
 
+        $consumer = escapeshellarg($this->config->getConsumer());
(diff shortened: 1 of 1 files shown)


mefellows:
Thanks so much!

PR #184, opened 2020-12-13
feat: Moved to GitHub Actions

### LINKED ISSUE #183: Chore: migrate away from TravisCI
TravisCI (org) is dropping support for OSS builds at the end of this month, meaning they must be transferred to travis-ci.com, which uses a credit system. Builds may fail (and already are) if the number of credits is exhausted within the month.

We are recommending switching to GH Actions, but the choice of CI provider is up to the project maintainer.

See also:

https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom
https://travis-ci.community/t/org-com-migration-unexpectedly-comes-with-a-plan-change-for-oss-what-exactly-is-the-new-deal/10567/5

siad007:
Removed travis and appveyor in favor of github actions.

@cfmack @mefellows should I delete the config files also?
AFAIS the new actions are not auto triggered. Some configuration on project side maybe?
Fork was autotriggered successful https://github.com/siad007/pact-php/actions

Closes #183


mefellows:
Fantastic - thanks!

Let's see about a transitional phase with GH Actions and when we're happy, we can cut over.

As for the triggers, I think it should just work - probably because it's a branch and not yet in master?

mefellows:
This could probably just be brought straight in and iterated on (because it should have no effect on the current build etc.), but I'll leave it with @cfmack to decide as the repo maintainer.

PR #192, opened 2021-02-02
fix type hints

### LINKED ISSUE #191: typehints use backslashes to fully qualify `string` types
references starting with a backslash indicate classes or functions. built in scalar types are reserved keywords and not namespace bound.

```php
<?php

namespace App;

class Test {
    private \string $hello;
}
```

> Fatal error: Type declaration 'string' must be unqualified in /in/WS8PP on line 6

for reference: https://github.com/phpstan/phpstan/issues/540

it is causing issues here:
- https://github.com/pact-foundation/pact-php/blob/6182c4f5c684c88959acba80abba08e2e0449480/src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfigInterface.php#L153
- https://github.com/pact-foundation/pact-php/blob/6182c4f5c684c88959acba80abba08e2e0449480/src/PhpPact/Standalone/ProviderVerifier/Model/VerifierConfigInterface.php#L158

ctrl-f5:
fixes #191 type hints based on phpstan level 5 analysis

### DIFF
--- a/src/PhpPact/Consumer/Hook/ContractDownloader.php
+++ b/src/PhpPact/Consumer/Hook/ContractDownloader.php
@@ -5,6 +5,8 @@ namespace PhpPact\Consumer\Hook;
 use Exception;
 use PhpPact\Http\ClientInterface;
 use PhpPact\Http\GuzzleClient;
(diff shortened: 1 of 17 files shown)


cfmack:
Thank you!

Issue #193, opened 2021-02-02
phpunit versions should be constrained through composer.json


ctrl-f5:
phpunit should be moved from `require-dev`to `require` so this package can constrain which versions of phpunit it supports.

Currently the support is documented in the readme file, but one can install it with unsupported versions. by moving phpunit into `require` composer can check supported versions during install.

YOU54F:
Completed by

 https://github.com/pact-foundation/pact-php/commit/7b50c96f820ed0f5cc422fb4c417a2df7b86f6d0

PR #194, opened 2021-02-02
Add phpstan for static analysis

ctrl-f5:
- add phpstan checks in the pipeline, level 5
- add phpunit to `require` so we can validate version (issue #193)
  * set minimum required phpunit version to 8.2.3 (https://github.com/sebastianbergmann/phpunit/issues/3728)
- fix static analysis issues
- extends and contains PR https://github.com/pact-foundation/pact-php/pull/192

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         php: [ '7.3', '7.4' ]
-    
+
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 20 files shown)


ctrl-f5:
github actions seem very unstable, composer install randomly fails with "access denied" in the file system or "resource unavailable"

cfmack:
Acknowledged that issue is randomly failing.  Thanks for finding the link.   We moved to Github Actions after Travis changed their terms of use.

PR #197, opened 2021-04-26
feat: Updated dependencies - php 8.0 compat

siad007:
With updated composer dependencies it will also work with windows now.
Only
PHP 8.0 on windows-latest with lowest dependencies
still fails.
I will investigate the next days here.


jildertmiedema:
Hi did a small addition to your work. The composer stuff now works on windows lowest-deps. I've created an extra PR (https://github.com/pact-foundation/pact-php/pull/202) however i am not allowed to run github actions in this repo.

Maybe it can help you to get this running.

PR #200, opened 2021-06-08
feat: Support matchers in headers

jildertmiedema:
The `string` was to strict and prevented the headers from being matchable.

With this fix you can use matcher in headers like this:
```php
$request = new ConsumerRequest()
    ->setMethod('POST')
    ->setPath('/api/call')
    ->addHeader('Content-Type', 'application/json')
    ->addHeader('X-Auth-Token', $matcher->like('auth-token'));
```


cfmack:
Merging per code review

PR #205, opened 2021-07-01
fix: Fixing pipeline issues with PHP 8 compatibility

Sertion:
This should fix the issues with the pipeline not finishing.

This pipeline here will still use the GitHub Actions from the main branch but [you can see them working in my fork](https://github.com/Sertion/pact-php/actions/runs/990338551).

I have fixed a few things here but I believe the main issue to be that we mixed different major versions of the amp-packages.


Sertion:
Unless provided feedback requires any changes I feel my intended change here is complete. Will someone be able to review it?

cfmack:
The PHP 7.3 and 7.4 are still failing on static code analysis, at least at quick glance.

PR #209, opened 2021-08-16
feat: reduce composer/semver required version

WaylandAce:
This change would improve package compatibility.


cfmack:
Resolved in #214

PR #211, opened 2021-08-22
style: format annotation

tienvx:


cfmack:
Completed as part of #213

PR #217, opened 2021-09-07
fix: GuzzleHttp\Psr7\Uri invalid path modification

### LINKED ISSUE #207: InvalidArgumentException: The path of a URI with an authority must start with a slash "/" or be empty
Trying to run tests against provider I've received the following issue:

```
InvalidArgumentException: The path of a URI with an authority must start with a slash "/" or be empty

app/vendor/guzzlehttp/psr7/src/Uri.php:729
app/vendor/guzzlehttp/psr7/src/Uri.php:492
app/vendor/pact-foundation/pact-php/src/PhpPact/Standalone/ProviderVerifier/Verifier.php:155
app/tests/Contract/VerifierTest.php:113
```

It seems that this problem is caused by guzzle. See this: https://github.com/guzzle/psr7/blob/master/CHANGELOG.md?plain=1#L183

So I assume the fix will be to add leading slash to this line: https://github.com/pact-foundation/pact-php/blob/master/src/PhpPact/Standalone/ProviderVerifier/Verifier.php#L145

nbayramberdiyev:
This PR fixes `InvalidArgumentException` exception (with message 'The path of a URI with an authority must start with a slash "/" or be empty') during provider verification from pact broker.

E.g. `(new \GuzzleHttp\Psr7\Uri)->withPath('foo/bar')` throws an exception because the path of a URI with an authority must start with a slash "/" or be empty (See [change log](https://github.com/guzzle/psr7/blob/master/CHANGELOG.md#changed-5)).

Fixes #207

### DIFF
--- a/src/PhpPact/Standalone/ProviderVerifier/Verifier.php
+++ b/src/PhpPact/Standalone/ProviderVerifier/Verifier.php
@@ -152,7 +152,7 @@ class Verifier
      */
     public function verify(string $consumerName, string $tag = null, string $consumerVersion = null): self
     {
-        $path = "pacts/provider/{$this->config->getProviderName()}/consumer/{$consumerName}/";
+        $path = "/pacts/provider/{$this->config->getProviderName()}/consumer/{$consumerName}/";
 
         if ($tag) {
             $path .= "latest/{$tag}/";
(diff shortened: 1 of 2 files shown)


nbayramberdiyev:
Also seems related to #57

cfmack:
Thanks!

PR #228, opened 2021-12-08
feat: Add installer for system pact binaries if preinstalled on posix systems

andytson-inviqa:
Its necessary to use a different ruby vendor for arm64 architectures, so this gives the ability to use gem packages pre-installed on a system ruby, such as docker `linux/arm64` images when `linux/amd64` emulation is too unstable


mefellows:
Nice idea!

cfmack:
Github actions was acting silly.  Tests re-ran and all passed on https://github.com/pact-foundation/pact-php/actions/runs/1811919934

PR #230, opened 2022-01-14
fix: Address PHP 8.1 deprecations

agustingomes:
Most of them were related to the change of type of PHP Core interfaces, and to maintain compatibility with PHP 7.3, the attribute `#[\ReturnTypeWillChange]` is used, which before PHP 8.0 is interpreted as a comment.

Another change is related with a deprecation when passing `null` in the 3rd argument of `str_replace`.


mefellows:
I've just unblocked the build, thanks for this.

I'll leave it to @cfmack to review/merge in if applicable.

cfmack:
This is great @malukenho .  Thank you.

PR #232, opened 2022-02-16
fix: deprecation of null being passed to "trim()"

malukenho:


malukenho:
@cfmack is it possible to release a version with the master state?

cfmack:
Released 7.1.1

Issue #231, opened 2022-01-21
BrokerHttpClientInterface::publishJson has different order of parameters in interface and implementation


mkluczka-cobiro:
BrokerHttpClient
```
    public function publishJson(string $json, string $version)
    {
        $array    = \json_decode($json, true);
```

BrokerHttpClientInterface
```
    public function publishJson(string $version, string $json);
```

malukenho:
@cfmack it was addressed in #233.
It can be closed.

YOU54F:
thanks for resolving @malukenho

PR #233, opened 2022-02-23
fix: parameter order mismatch

malukenho:
Linking #231


EHER (approved):

malukenho:
It looks like windows ci is failing randomly. I had to trigger it 3 times. /cc @cfmack

EHER (approved):

PR #235, opened 2022-03-03
feat: Support all variations of ISO 8601 time zone designators

mdio:
Especially important for us is the support of the +/-hhmm format  (e.g. `2015-08-06T16:53:10.123+0100`) as it's used in our Java services.


cfmack (approved):

mdio:
Thank you for the swift review & merge!

PR #239, opened 2022-03-24
Open the PactTestListener

jildertmiedema:
It is hard to extend the PactTestListener therefore i have made the fields "protected"

This way we can start our own `MockServer` using a dockerized version of the ruby standone version


jildertmiedema:
> Why not just create your listener implementing the interface?

The interface is a generic phpunit interface. The most important function in the class `PactTestListener` is `endTestSuite` and then i have to copy that code. And it will get out of sync some day.

PR #240, opened 2022-04-03
feat: Add support for publishing verification results with a branch

### LINKED ISSUE #223: Add support for publishing verification results with a branch using the pact-ruby-standalone
The Pact Broker now supports branches as first class entities. You can read more about this here: https://github.com/pact-foundation/docs.pact.io/blob/feat/deployments-and-releases/website/blog/2021-07-04-why-we-are-getting-rid-of-tags.md

To allow users to use this feature:

* Upgrade the pact-ruby-standalone to at least version `1.88.70`
* Add an appropriately prefixed "branch" property to the user facing interface (make it match the existing provider name, tag, and version properties)
* Pass the branch through to the pact provider verifier CLI using the parameter `--provider-version-branch`

WaylandAce:
The Pact Broker now supports branches as first class entities. You can read more about this here: https://github.com/pact-foundation/docs.pact.io/blob/feat/deployments-and-releases/website/blog/2021-07-04-why-we-are-getting-rid-of-tags.md

Closes #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,7 @@ $config = new VerifierConfig();
 $config
     ->setProviderName('someProvider') // Providers name to fetch.
     ->setProviderVersion('1.0.0') // Providers version.
+    ->setProviderBranch('main') // Providers git branch name.
     ->setProviderBaseUrl(new Uri('http://localhost:58000')) // URL of the Provider.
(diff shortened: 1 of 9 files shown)


cfmack:
Thanks @WaylandAce

PR #241, opened 2022-04-03
feat: support "Publish Contracts" endpoint

WaylandAce:
This is the preferred endpoint with which to publish contracts (previously, contracts were published using multiple calls to different endpoints to create the tag and contract resources)

### DIFF
--- a/example/phpunit.consumer.xml
+++ b/example/phpunit.consumer.xml
@@ -22,6 +22,7 @@
         <env name="PACT_CONSUMER_NAME" value="someConsumer"/>
         <env name="PACT_CONSUMER_VERSION" value="1.0.0"/>
         <env name="PACT_CONSUMER_TAG" value="master"/>
+        <env name="PACT_CONSUMER_BRANCH" value="master"/>
         <env name="PACT_PROVIDER_NAME" value="someProvider"/>
         <env name="PACT_OUTPUT_DIR" value=".\example\output"/>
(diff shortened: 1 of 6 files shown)


bethesque:
Awesome! Can this print out the text from the notices please? Extra points if it can be colourized (as per [this logic](https://github.com/pact-foundation/pact_broker-client/blob/2a51e37f373841c673a947881dded21c4b7de70b/lib/pact_broker/client/colorize_notices.rb#L21)), but just printing out the text is fine. You can see an example of the output here:

https://github.com/pact-foundation/pact_broker-client/runs/5546727028?check_suite_focus=true#step:5:265

Screen Shot 2022-04-04 at 9 46 05 am

bethesque:
Actually Pavel, this version of the php Pact library already has the Pact CLI embedded in it. We're actually trying to get as many of the points that interact with the Pact Broker to go through the shared Pact CLI as they can, rather than using native code, because it keeps the surface area for updates as small as possible for rolling out changes. I know it seems like a small amount of code, but when you have 10+ languages repeating the same thing over and over, rolling out improvements to the end users becomes a very slow process. For example, it's taken us almost 2 years to roll out branches and environments!

If you're able to make this code call the [publish](https://docs.pact.io/pact_broker/client_cli/readme#publish) command of the CLI, that would be the preferred solution.

WaylandAce:
https://github.com/pact-foundation/pact-php/pull/243

PR #244, opened 2022-04-21
feat: pass pact-broker-base-url argument to verifier process

WaylandAce:
With this change it's possible to fetch pacts directly from broker.

### DIFF
--- a/src/PhpPact/Standalone/ProviderVerifier/Verifier.php
+++ b/src/PhpPact/Standalone/ProviderVerifier/Verifier.php
@@ -144,7 +144,7 @@ class Verifier
         }
 
         if ($this->config->getBrokerUri() !== null) {
-            $parameters[] = "--pact-broker-base-url={$this->config->getBrokerUri()}";
+            $parameters[] = "--pact-broker-base-url={$this->config->getBrokerUri()->__toString()}";
         }
 
         return $parameters;
--- a/src/PhpPact/Standalone/ProviderVerifier/Verifier.php
+++ b/src/PhpPact/Standalone/ProviderVerifier/Verifier.php
@@ -143,6 +143,10 @@ class Verifier
             $parameters[] = "--include-wip-pacts-since={$this->config->getIncludeWipPactSince()}";
         }
 
+        if ($this->config->getBrokerUri() !== null) {
+            $parameters[] = "--pact-broker-base-url={$this->config->getBrokerUri()}";
+        }
+
         return $parameters;
     }
 
--- a/tests/PhpPact/Standalone/ProviderVerifier/VerifierTest.php
+++ b/tests/PhpPact/Standalone/ProviderVerifier/VerifierTest.php
@@ -36,6 +36,7 @@ class VerifierTest extends TestCase
             ->setBrokerToken('someToken')
             ->setBrokerUsername('someusername')
             ->setBrokerPassword('somepassword')
+            ->setBrokerUri(new Uri('https://example.broker/'))
             ->addCustomProviderHeader('key1', 'value1')
             ->addCustomProviderHeader('key2', 'value2')
             ->setVerbose(true)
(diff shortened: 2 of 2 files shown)


mefellows:
I'm confused - what was the situation prior to this? I'd imagine 99.9% of the use case _is_ fetching from a broker. I can see a bunch of other broker related flags, so I'm wondering if I'm missing something obvious about how that was specified prior to this change?

WaylandAce:
@mefellows 
All previous cases of verify:
1) Fetch by tag (/pacts/provider/$providerName}/consumer/$consumerName/)
2) By files
3) Verify all (/pacts/provider/$providerName/latest)
4) By tag (/pacts/provider/$providerName/latest/$tag)
5) By config (VerifierConfig).

This PR improves 5 option. Actually, pact broker url can be set thru PACT_BROKER_BASE_URL env variable. But env variable is not related to VerifierConfig. With this PR behavior of 5th verification is more clear. 


```
      [--pact-broker-base-url=PACT_BROKER_BASE_URL]                          # Base URL of the Pact Broker from which to retrieve the pacts. Can also be set using the environment variable PACT_BROKER_BASE_URL.

```

YOU54F:
We mentioned this option in the docs before, but it doesn’t look like it was being mapped to the ruby verifier option as added here. 🧐

https://github.com/pact-foundation/pact-php#provider-verification

Is that just an omission that this pr will resolve?

WaylandAce:
Right now this option is used to publish pacts only, not fetch.

YOU54F:
Ok thanks dude, so is that just an omission in the docs, or do we need to add anything additional in

I note you've made a couple of other changes, so I imaging you are actively using this, so appreciate the contributions, and I am sure @cfmack does as well.

Will look to get the pulled in as soon as possible (just trigged the builds again, as soon we failing, more than likely unrelated)

Just doing some OSS housekeeping across various repos today, and I've not looked at Pact-php much

YOU54F:
This test is failing https://github.com/pact-foundation/pact-php/blob/3915cde524346fd7207d20dad6cb2ce1268c831a/tests/PhpPact/Standalone/ProviderVerifier/VerifierTest.php#L113

see build output https://github.com/pact-foundation/pact-php/runs/6159378723?check_suite_focus=true

```
[2022-04-25 14:29:06] console.ERROR: err >  [] []
Test 'PhpPactTest\Standalone\ProviderVerifier\VerifierProcessTest::testRunForwardsException' ended
Test 'PhpPact\Standalone\ProviderVerifier\VerifierTest::testGetArguments' started
Test 'PhpPact\Standalone\ProviderVerifier\VerifierTest::testGetArguments' ended
Test 'PhpPact\Standalone\ProviderVerifier\VerifierTest::testGetArgumentsEmptyConfig' started
Test 'PhpPact\Standalone\ProviderVerifier\VerifierTest::testGetArgumentsEmptyConfig' ended
Test 'PhpPact\Standalone\ProviderVerifier\VerifierTest::testBuildValidPathToPactBroker with data set #0 ('someProviderName', 'someProviderName', null, '11111', '/pacts/provider/someProviderN...11111/')' started
PHP Fatal error:  Method Mock_Uri_3c0c9044::__toString() must not throw an exception, caught PHPUnit\Framework\ExpectationFailedException: GuzzleHttp\Psr7\Uri::__toString(): string was not expected to be called more than once. in /home/runner/work/pact-php/pact-php/src/PhpPact/Standalone/ProviderVerifier/Verifier.php on line 0
Script phpunit --debug -c example/phpunit.all.xml handling the test event returned with error code 255
```

It looks like the uri is toString'd here, possibly why its complaining

https://github.com/pact-foundation/pact-php/blob/3915cde524346fd7207d20dad6cb2ce1268c831a/src/PhpPact/Standalone/ProviderVerifier/Verifier.php#L177-L179

cfmack:
Let me know when the tests are fixed.  Happy to help contribute.

YOU54F:
Thanks @WaylandAce ,

@cfmack are you okay to merge and release when you are happy? Cheers all

PR #249, opened 2022-05-24
Bugfix/update pact url

Lewiscowles1986:
This should fix master builds. At least it seems to in my local

Command

```
git clone https://github.com/Lewiscowles1986/pact-php
cd pact-php
git checkout bugfix/update-pact-url
composer install
php vendor/bin/phpunit tests\\PhpPact\\Standalone\\Broker\\BrokerTest.php
```

## Notes:

* I Changed the formatting of some of the file to make it easier to step-through in a debugger and find the problem. While concise the prior code was not helping me to find the problem.

* It reads like the config & broker initialization could be delegated to a short-hand function `withBroker()`. What do you think about this?

### DIFF
--- a/tests/PhpPact/Standalone/Broker/BrokerTest.php
+++ b/tests/PhpPact/Standalone/Broker/BrokerTest.php
@@ -47,13 +47,16 @@ class BrokerTest extends TestCase
      */
     public function describeVersion(): void
     {
-        $this->assertArrayHasKey('number', (new Broker(
-            (new BrokerConfig())
-                ->setPacticipant(\rawurlencode('Animal Profile Service'))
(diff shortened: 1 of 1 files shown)


YOU54F:
Lets give this a go, and see what happens. It doesn't look like it will trigger a release automatically off master, as the last release tag is quite from Aug 21

https://github.com/pact-foundation/pact-php/releases

Hmm actually, the latest is 7.1.2 https://packagist.org/packages/pact-foundation/pact-php#7.1.2

with the timestamp `2022-04-25 20:45 UTC`

It was tagged, but just not added to release, I note we don't have a `CHANGELOG` either.

PR #253, opened 2022-05-27
Suggestion: Build on schedule once per month

Lewiscowles1986:


YOU54F:
Nice thanks buddy, any chance you could 

1. force push with a re-written commit message `chore: xyz`
2. add a comment next to the cron, just so it is easy to grok for us non computers :) or just add a link to crontab guru https://crontab.guru/#0_6_1

Lewiscowles1986:
In the editor for GitHub it has a hover explanation. Sadly not the viewer, only the editor 😆 .

image
https://github.com/Lewiscowles1986/pact-php/edit/suggestion/build-on-schedule-once-per-month/.github/workflows/build.yml?pr=%2Fpact-foundation%2Fpact-php%2Fpull%2F253

YOU54F:
Yes please https://github.com/pact-foundation/pact-php/pull/253#issuecomment-1139626738 not everyone will be editing via the GUI (that is a sweet touch tho), some still prefer vim/emacs, and I use vscode, there is probably an extension if it doesn't natively show it.

Lewiscowles1986:
Updated so that the comment is there btw @YOU54F

YOU54F:
nice one fella!

Issue #245, opened 2022-05-09
Error filtering ->verify($consumerName) after 7.1.2 update


rubengg86:
After this change, https://github.com/pact-foundation/pact-php/commit/3915cde524346fd7207d20dad6cb2ce1268c831a, if we try to filter the verifier by consumer name using:

` $verifier = (new MessageVerifier($config))
            ->setCallbacks($callbacks)
            ->verify($this->consumerService);`

Doesn't work because is trying to verify with all pacts in the broker. If we rollback to 7.1.0 works fine

cfmack:
@rubengg86, happy to take a PR.  Maybe from @WaylandAce

WaylandAce:
I'll take a look

YOU54F:
Looks at that commit, your issue sounds unrelated

PR #261, opened 2022-05-31
Consumer selection

jildertmiedema:
We had a problem with consumer filtering. Like https://github.com/pact-foundation/pact-php/issues/245. This is how we fixed it.

Not sure if it breaks other stuff.

### DIFF
--- a/src/PhpPact/Standalone/ProviderVerifier/Verifier.php
+++ b/src/PhpPact/Standalone/ProviderVerifier/Verifier.php
@@ -176,6 +176,9 @@ class Verifier
 
         $uri = $this->config->getBrokerUri()->withPath($path);
 
+        $filter = ['consumer' => $consumerName, 'latest' => true];
+        $this->config->getConsumerVersionSelectors()->addSelector(json_encode($filter));
+
         $arguments = \array_merge([$uri->__toString()], $this->getArguments());
 
         $this->verifyAction($arguments);


Lewiscowles1986:
Quick thought: Any chance of a test to guard against regression?

cfmack:
Tests did not fail.  However, I think we want to be able to set latest to false somewhere.

YOU54F:
It should be opt-in by user not implicit, without an option to disable.

You should also add a test to cover your change

https://docs.pact.io/pact_broker/advanced_topics/consumer_version_selectors#advanced

```
{ "tag": "<tag>" } - all versions with the specified tag
{ "tag": "<tag>", "latest": true }- the latest version for each consumer with the specified tag
```

This would change user facing behaviour where they previously fetched all pacts, by tag, but now only fetch the latest.

tienvx:
Please change the target branch to `release/9.x`. I'm sorry about that.

YOU54F:
Hey,

I am going to close down this PR, I apologise but I also don't think it is the right approach

> It should be opt-in by user not implicit, without an option to disable.

As an end user today, you can pass in your own selectors to solve use cases which will vary over time, and we should not provide filtering on the users behalf without them knowing.

Happy to reconsider, but I would say it would be worth describing your workflow and the issues you are facing, rather than just providing the solution, as in many cases there are probably better solutions (such as looking at your pact verification workflow and upgrading it to use named branches/environment selectors, over I assume tags) 

Thanks for proposing a change and hope this doesn't put you off in the future!

PR #264, opened 2022-07-28
feat: Support matchers in path

turb0bur:
Added matchers support in the ConsumerRequest path in order to use path variables

### DIFF
--- a/src/PhpPact/Consumer/Model/ConsumerRequest.php
+++ b/src/PhpPact/Consumer/Model/ConsumerRequest.php
@@ -14,7 +14,7 @@ class ConsumerRequest implements \JsonSerializable
     private $method;
 
     /**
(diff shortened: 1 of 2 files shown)


Lewiscowles1986:
I Like this 👍

cfmack:
I kicked off the CI.  Lint checks failed: [7.3 Lint](https://github.com/pact-foundation/pact-php/runs/7599360533?check_suite_focus=true)

Lewiscowles1986:
Notice it doesn't give any helpful errors about how that file fails lint...

On second glance, this does widen a lot of interfaces; which might be problematic for consumers. Might be worth giving it a go-over with phpstan and running the unit suite.

cfmack:
Closed in #273

PR #257, opened 2022-05-28
Collab with tienvx: use ffi

Lewiscowles1986:
* https://github.com/pact-foundation/pact-php/issues/258
* https://github.com/pact-foundation/pact-php/pull/210

PR #267, opened 2022-09-04
php-cs-fixer

mmoll:
Fixes the first step of the pipeline.

PR #268, opened 2022-09-04
fix: broken animal profile service URL

Lewiscowles1986:
Honestly I'm so mad I'm needing to sumbit another PR to an API testing product that isn't managing it's tests or API's.

Please put in place some process. Earlier this year it was changed credentials (which are hard-coded); now it's a rename of a service that this downstream relies on.

I Don't care if this merges as I don't use pactflow. Honestly every interaction upsets me because of this basic, preventable nonsense. This doesn't need to merge if someone renames `Animal Profile Service V3` back to `Animal Profile Service`.

Maybe a better pattern might be having a canned broker that lives in this repo, so does not need credentials, and won't flake based on network level changes. You never want to try to contribute and have network resource effects give false positives.

The actual test seems to be checking the broker command has received good arguments; so could a `--test` flag be added to sub-processes, so that they don't actually hit the network?

This was causing flakes in updating https://github.com/pact-foundation/pact-php/pull/257

### DIFF
--- a/tests/PhpPact/Standalone/Broker/BrokerTest.php
+++ b/tests/PhpPact/Standalone/Broker/BrokerTest.php
@@ -48,7 +48,7 @@ class BrokerTest extends TestCase
     public function describeVersion(): void
     {
         $config = new BrokerConfig();
-        $config->setPacticipant(\rawurlencode('Animal Profile Service'))
+        $config->setPacticipant(\rawurlencode('Animal Profile Service V3'))
             ->setBrokerUri(new Uri('https://test.pactflow.io'))
             ->setBrokerUsername('dXfltyFMgNOFZAxr8io9wJ37iUpY42M')
             ->setBrokerPassword('O5AIZWxelWbLvqMd8PkAVycBJh2Psyg1');
@@ -67,7 +67,7 @@ class BrokerTest extends TestCase
     public function listLatestPactVersions(): void
     {
         $config = new BrokerConfig();
-        $config->setPacticipant(\rawurlencode('Animal Profile Service'))
+        $config->setPacticipant(\rawurlencode('Animal Profile Service V3'))
             ->setBrokerUri(new Uri('https://test.pactflow.io'))
             ->setBrokerUsername('dXfltyFMgNOFZAxr8io9wJ37iUpY42M')
             ->setBrokerPassword('O5AIZWxelWbLvqMd8PkAVycBJh2Psyg1');


Lewiscowles1986:
Looks like https://github.com/pact-foundation/pact-php/pull/267 is needed for this to merge.

mmoll (approved):
can confirm that this fixes the problem for me locally.

mefellows:
Yes, the issue is because the build is relying on a pacticipant Animal Profile Service that doesn’t exist. This isn’t a broken URL in Pactflow, it’s a bad test that assumes an application should be there when it can be deleted by anyone (it’s a public broker!). 

> Maybe a better pattern might be having a canned broker that lives in this repo, so does not need credentials, and won't flake based on network level changes. You never want to try to contribute and have network resource effects give false positives.

That's reasonable, I think pact-python does a similar thing. There are different schools of thought, and ultimately it's up to the maintainer to choose how to run the builds and if a broker is a necessary thing to have as part of the process.

This looks to be an integration test for invoking the CLI, so I think a local broker is probably a better substitute here, unlikely an e2e test that might want to test key scenarios.

We have some examples that use Pactflow in other projects, but we always setup the consumer, publish the consumer and then verify the provider so we can be sure the data is there.

TimothyJones:
> Honestly I'm so mad I'm needing to sumbit another PR to an API testing product that isn't managing it's tests or API's.

Ex-pact-js maintainer here (I still contribute occasionally but don't look after any repository directly any more). Thanks for the PR, even though it was reluctantly submitted.

It's worth pointing out that Pact is an open source product largely maintained by volunteers - even though there's a close relationship with Pactflow (who provide hosted broker services - and have generously donated a test account for the projects to use in CI), these tests are still community maintained.

There are lots of ways that the project's process can be improved, and I think your feedback is fair. One of the strengths of Pact (and something that I think is pretty unique for an API testing project) is that it runs in multiple languages and frameworks. It's true that Pact PHP hasn't had the same love as some of the other projects - in large part because it doesn't get as many community submissions and as far as I know is generally used less than the JVM / JS libraries.

Pact is an ambitious project, with only a small team - and even though it has a few areas that need cleanup, I'm not aware of anything else that is as effective and easy for doing API contract testing.

> Maybe a better pattern might be having a canned broker that lives in this repo, so does not need credentials, and won't flake based on network level changes. You never want to try to contribute and have network resource effects give false positives.

This is a good suggestion. Some of the tests in Pact JS used to work this way - but they got out of sync with the actual broker so stopped providing value. When I was a maintainer, I nearly rewrote those as pacts (this would avoid the sync issue, and frankly I don't know why they weren't pacts to start with) - but a challenge is that then the examples would be extra confusing. It's helpful to have the examples be as close to real world usage as possible - and for that reason, I think it's worth taking on the risk of the occasional failure due to network issues (this is pretty rare, or at least it was while I was a maintainer).

> this basic, preventable nonsense

Fixes for preventable nonsense are always welcome. In this PR, you've fixed something, highlighted a different problem, and proposed a solve. I'm not a fan of raising problems without solutions- so thanks, even if this is your only PR, it's appreciated.

And, if you want to contribute more, I know it would be welcome. The Pact (and Pactflow) teams are certainly the nicest and most dedicated OSS maintainers I've ever worked with.

TimothyJones:
> We have some examples that use Pactflow in other projects, but we always setup the consumer, publish the consumer and then verify the provider so we can be sure the data is there.

I created #269 to track improving the examples in line with this suggestion - regardless of whether a mock broker is used, that's a clear improvement.

YOU54F:
@Lewiscowles1986 appreciate your frustrations, however there are far better ways to get maintainers on side.

A little bit of politeness goes a long way. 

> It's worth pointing out that Pact is an open source product largely maintained by volunteers - even though there's a close relationship with Pactflow (who provide hosted broker services - and have generously donated a test account for the projects to use in CI), these tests are still community maintained.

We appreciate your pull requests in the past, in helping improve Pact-PHP in recent months and your current work. If you look at the commit history, there are three core contributors to Pact-PHP, who have done this voluntarily.

We rely on community contributions, and with there being a large surface area and a small number of people, it is difficult to prioritise and implement good engineering practises across the board and ensure everything is covered. That doesn't mean that we don't want to! I totally agree that a testing tool, should have excellent testing practises, and there is significant room for improvement, but we are slowly chipping away. 

> This doesn't need to merge if someone renames Animal Profile Service V3 back to Animal Profile Service.

It wasn't renamed (The V3 refers to a copy of the original service, for a Pact JS V3 example)

All Pacts were deleted from the public broker (the test broker was cleared two weeks ago ) and there hasn't been a run which has uploaded the required contracts for this test. If this test did need to call out to the network, it should ideally be setting up any required data, rather than relying on it being there in the first place.

Will see if we can unblock you for the time being and we will take your feedback on board.

YOU54F:
A rename was done 19 months ago, with the provider coming from the pact-js project

https://github.com/pact-foundation/pact-js/commit/9270ca9ab81ac9c71ffaf25ffda4bcdb487bce27#diff-18d291b57f7eb1321ff01436695e50c8c110dc150504a85a07f23a6be01979c4

I've just re-uploaded a Pact file for that consumer/provider pair

Lewiscowles1986:
Guessing this means master is now working and I can close this. So I will

PR #270, opened 2022-09-10
Rust FFI, try 3

mmoll:


mmoll:
tests seem super flahey... Seems to work on Linux/arm64, no idea about MAC...

Lewiscowles1986:
This is great @mmoll ❤️

cfmack:
Thank you, @mmoll and @Lewiscowles1986 .  I am happy to merge it if it passes but here's where I sit on the matter.  As you can see like @YOU54F stated, certainly true on the PHP side, there is only contributor support like the hard work you just did.   I have little bandwidth for extensive changes like I did when another dev and I ported Pact.Net many years ago.   Hence, if this dramatically changes the underlying functionality, those issues will likely linger until folks like you two can assist with.

YOU54F:
Thanks all! I’m happy to support with the testing when it gets released ( not sure if we want to do a beta release )

And will add it to my list to look at the pipeline flakiness, as that is a big pain for contributors and Eros confidence. It happens in other repos as well so pact-php isn’t the outlier

Thank you everyone!

YOU54F:
Sorry for the delay @mmoll - great work on this.

Got a chance to look at this today, dropped a commit with some comments in the code, based on some initial observations.

https://github.com/pact-foundation/pact-php/commit/e1e0e3dc9d76ecf2871b097baa8b3e6fda5ebfc0#diff-5919afd90702d51e005580af895ec9df6127dae6c92748063b2bf8190fadec7aR61

Changes are on my fork at the moment

https://github.com/YOU54F/pact-php/commit/e1e0e3dc9d76ecf2871b097baa8b3e6fda5ebfc0

PR #272, opened 2022-09-14
fix: provider-verifier bad arguments naming

### LINKED ISSUE #271: Invalid log directory option passed to the pact-provider-verifier cli
In case we define a log directory to the `VerifierConfig`, [the `Verifier` will append a `--log` option](https://github.com/pact-foundation/pact-php/blob/master/src/PhpPact/Standalone/ProviderVerifier/Verifier.php#L131) to the pact-provider-verifier command. 

However, according to [pact-provider-verifier documentation](https://github.com/pact-foundation/pact-provider-verifier#usage), the option is named `--log-dir`

Veridis:
closes #271 

I've also change the `--verbose` argument to `--verbose=VERBOSE`

### DIFF
--- a/src/PhpPact/Standalone/ProviderVerifier/Verifier.php
+++ b/src/PhpPact/Standalone/ProviderVerifier/Verifier.php
@@ -124,11 +124,11 @@ class Verifier
         }
 
         if ($this->config->isVerbose() === true) {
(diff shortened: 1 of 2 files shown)


cfmack:
Thanks @Veridis

PR #273, opened 2022-09-15
Feature/support matchers

cfmack:
Running tests for  #264


cfmack:
As all the tests pass here, after merging from master, I'll merge here in place of #264

PR #210, opened 2021-08-21
feat: Use Rust FFI

tienvx:

PR #278, opened 2022-12-14
Use Rust FFI #1: Drop PHP 7.4, test PHP 8.2

tienvx:
This is my first pull request in a series of "small" pull requests I split from https://github.com/pact-foundation/pact-php/pull/210


.github/workflows/build.yml
```diff
@@ -47,30 +47,27 @@ jobs:
...
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
+          extensions: openssl, sockets, curl, zip
           php-version: ${{ matrix.php }}
 
-      - name: Copy the inis
-        if: runner.os == 'Windows'
-        run: Copy-Item -Path .\tests\PhpPact\php.ini -Destination C:\tools\php\ -Force
```
tienvx, line 70: I'm not sure what is this fix for. But I think it's not needed in PHP ^8.0

.github/workflows/build.yml
```diff
@@ -13,23 +13,23 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [ '7.4', '8.0', '8.1' ]
+        php: [ '8.0', '8.1' ]
```
tienvx, line 16: This doesn't include PHP 8.2 because PHP CS FIXER doesn't have a new release that support PHP 8.2 yet.

Do we need to check coding standards for every PHP versions?

How about only run it on latest supported PHP version only? In this case only PHP 8.1?

Lewiscowles1986 (approved):
Looks good to me.

Issue #276, opened 2022-11-06
Install the pact binaries from composer instead of within the code


deanstalker:
It would be preferable to see a function called from the `composer install/update` flow which downloads and populates the vendor/bin folder instead of managing the install within the runtime code.

The alternative is to configure the destination directory so we can bake the binaries into a docker image and use our paths instead of the hard-coded reference to the vendor folder. We don't allow web users to write to the vendor folder as a rule of thumb.

YOU54F:
Both of those make sense, it was raised by another member and addressed by Tien in the upcoming pact-rust (FFI) backed version.

Would definitely consider an interim change against the ruby backed version, as a new PR 

https://github.com/pact-foundation/pact-ruby-standalone/pull/100

means that we open support up to additional platforms (arm64 linux and osx, and 32 bit windows, which was previously dropped in a traveling ruby update, but I've brought it back)

YOU54F:
Ahh it's done, we just need to cut a new release, I've ran it through on my machine, that PR look good from @tienvx in #279

YOU54F:
you can test this out by using `dev-master` in your composer.json

PR #279, opened 2022-12-14
Use Rust FFI #2: Offload installers

tienvx:
This is 2nd pull request in a series of "small" pull requests I split from https://github.com/pact-foundation/pact-php/pull/210

This pull request depend on https://github.com/pact-foundation/pact-php/pull/278, please review it first.

There are number of reasons why I choose this approach:
* Update installers to downloads more files require adding quite a mount of code (see pull request above)
  * Need download at least 4 files
    * Pact Ruby Standalone
    * Pact FFI header
    * Pact FFI shared library
    * Pact stub server
  *  If we want supports plugins, 2 more for each plugin
      * csv
          * json meta file
          * executable file
      * protobuf
          * json meta file
          * executable file
  * Each file has different url/file-name/destination-path depend on OS and architecture.
  * Extracting `zip` or `tar.gz` file is easy, but extracting `gzip` file require more work, and we need to move the target file to the correct destination. `*.gz` don't restore the original file name.
  * Mark files as executable after downloading
  * Delete downloaded archived file after extracting
  * Determine correct path to scripts for each OS
* Maintain custom code does not look right to me
  * Download on runtime, not on running `composer install/update`
  * Need to re-download all files after installing/updating
  * No caching support

After looking for alternate solution, I found this library [civicrm/composer-downloads-plugin](https://github.com/civicrm/composer-downloads-plugin). But the author is also busy, he don't have time review my pull requests.

That's why I maintain my own fork. After making some changes, I released version 1.1, which is used by this pull request. This library has some features:
* Support all kind of files that composer suppport
* Download on `composer install/update` command
* Support caching from composer
* Don't need to re-download files if already downloaded
* Support variables for different url/path depend on OS and architecture.

I think this pull request also solve this ticket https://github.com/pact-foundation/pact-php/issues/276


composer.json
```diff
@@ -69,13 +70,29 @@
...
+                },
+                "url": "https://github.com/pact-foundation/pact-ruby-standalone/releases/download/v{$version}/pact-{$version}-{$os}{$architecture}.{$extension}",
+                "path": "bin/pact-ruby-standalone"
+            }
+        }
+    },
+    "config": {
+        "allow-plugins": {
+            "tienvx/composer-downloads-plugin": true
+        }
```
tienvx, line 96: Add this config to make running test easier. I think each project depend on this library also need to define this config again in `composer.json`

src/PhpPact/Standalone/Installer/Model/Scripts.php
```diff
@@ -4,101 +4,59 @@
...
     }
 
-    /**
-     * @return string
-     */
-    public function getPactMessage(): string
+    private static function getSuffix(): string
     {
-        return $this->pactMessage;
+        return (PHP_OS_FAMILY === 'Windows' ? '.bat' : '');
```
tienvx, line 60: If we use `symfony/process`, I think we don't have to do this.

We can discuss this again in the future pull request (maybe the 4th pull request)

script/create-pr-to-update-pact-ruby-standalone.sh
```diff
@@ -1,31 +0,0 @@
...
-done
-
-
-git add ${VERSION_FILES}
-git commit -m "feat(upgrade): update standalone to ${STANDALONE_VERSION}"
-git push --set-upstream origin ${BRANCH_NAME}
-
-hub pull-request --browse --message "feat: update standalone to ${STANDALONE_VERSION}"
-
-git checkout master
```
tienvx, line 31: Now we only need to update 1 number in `composer.json`. That's why I think this script is not needed.

src/PhpPact/Standalone/Installer/Model/Scripts.php
tienvx: https://github.com/pact-foundation/pact-php/pull/284

cfmack (approved):
Code looks good, I ran this locally and it passed, all tests in Github actions pass.

composer.json
YOU54F: I think if we can provide support for a user to support the ffi being provided by them in their own location (which should be fine as we can have to provide a path to the library loader) we can support users who dont want to, or cant enable scripts to download.

tienvx: I agree with it.

This feature can be useful for plugins. See this [link](https://pact-foundation.slack.com/archives/C047TCR7B6W/p1679881619953729) for explanation.

Currently I don't have any idea on how to do it. Can you create a ticket so we can look at it later?

YOU54F: of course buddy 👍🏾  shall do now

PR #282, opened 2023-01-02
Use Rust FFI #5: Provider

tienvx:
This is 5th pull request in a series of "small" pull requests I split from https://github.com/pact-foundation/pact-php/pull/210

This pull request depend on https://github.com/pact-foundation/pact-php/pull/281, please review it first.

PR #283, opened 2023-01-03
Use Rust FFI #6: Stub Server

tienvx:
This is 6th pull request in a series of "small" pull requests I split from https://github.com/pact-foundation/pact-php/pull/210

This pull request depend on https://github.com/pact-foundation/pact-php/pull/282, please review it first.


tests/PhpPact/Standalone/StubServer/StubServerTest.php
```diff
@@ -15,15 +15,13 @@ public function testStartAndStop()
...
             $port         = 7201;
             $endpoint     = 'test';
 
             $subject = (new StubServerConfig())
-                ->setPactLocation($pactLocation)
-                ->setHost($host)
+                ->setFiles($pactLocation)
                 ->setPort($port)
-            ->setEndpoint($endpoint);
+                ->setEndpoint($endpoint);
```
tienvx, line 24: `$endpoint` is not useful if we interact with `StubServer` directly like this. It's only useful when we interact with stub service via `StubServerHttpService`.

I think it's better if we move `$endpoint` to `StubServerHttpService::getJson` like this:

```php
// class StubServerHttpService
public function getJson(string $endpoint): string
    {
        $uri      = $this->config->getBaseUri()->withPath('/' . $endpoint);
        $response = $this->client->get($uri, [
            'headers' => [
                'Content-Type' => 'application/json',
            ],
        ]);

        return \json_encode(\json_decode($response->getBody()->getContents()));
    }
```

Do you think so?

tienvx: See https://github.com/pact-foundation/pact-php/pull/296

tienvx:

PR #280, opened 2022-12-26
Use Rust FFI #3: Consumer

tienvx:
This is 3nd pull request in a series of "small" pull requests I split from https://github.com/pact-foundation/pact-php/pull/210

This pull request depend on https://github.com/pact-foundation/pact-php/pull/279, please review it first.

PR #286, opened 2023-01-07
Add email and fromProviderState matchers

tienvx:
This pull request **does not** related to Rust FFI. I copy the matchers from [pact-js](https://github.com/pact-foundation/pact-js/blob/master/src/dsl/matchers.ts#L14) because I think it will be useful.

This pull request depend on https://github.com/pact-foundation/pact-php/pull/280, please review it first.


YOU54F:
yep email one is a goodie 👍🏾 I backported to the current implementation #308

PR #288, opened 2023-02-03
fix: display stderr on pact publish process failure

WaylandAce:
Display stderr on pact publish process failure

Unify behavior with https://github.com/pact-foundation/pact-php/blob/master/src/PhpPact/Standalone/ProviderVerifier/VerifierProcess.php#L65

### DIFF
--- a/src/PhpPact/Standalone/Broker/Broker.php
+++ b/src/PhpPact/Standalone/Broker/Broker.php
@@ -245,13 +245,18 @@ class Broker
                 $this->getArguments()
             )
         );
-        $runner->runBlocking();
 
-        if ($runner->getExitCode() !== 0) {
-            $this->logger->error($runner->getStderr());
+        try {
+            $runner->runBlocking();
+
+            $this->logger->info('out > ' . $runner->getOutput());
+            $this->logger->error('err > ' . $runner->getStderr());
+        } catch (\Exception $e) {
+            $this->logger->info('out > ' . $runner->getOutput());
+            $this->logger->error('err > ' . $runner->getStderr());
+
+            throw $e;
         }
-
-        $this->logger->debug($runner->getOutput());
     }
 
     public function testWebhook()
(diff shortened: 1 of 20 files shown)


Big-Shark (approved):

cfmack:
Simple enough (and thank you).   Looks like we are failing a Lint test @WaylandAce .   Can you help?  https://github.com/pact-foundation/pact-php/actions/runs/4086790206/jobs/7182694903

src/PhpPact/Standalone/Broker/Broker.php
```diff
@@ -245,13 +245,18 @@ public function publish(): void
...
 
-        $this->logger->debug($runner->getOutput());
+            $this->logger->info('out > ' . $runner->getOutput());
+            $this->logger->error('err > ' . $runner->getStderr());
+        } catch (\Exception $e) {
+            $this->logger->info('out > ' . $runner->getOutput());
+            $this->logger->error('err > ' . $runner->getStderr());
+
+            throw $e;
+        }
```
tienvx, line 259:
```suggestion
try {
            $runner->runBlocking();
} finally {
            $this->logger->info('out > ' . $runner->getOutput());
            $this->logger->error('err > ' . $runner->getStderr());
}
```
I think this code is shorter and work the same.

WaylandAce: But we need re-throw exception also, not just output

tienvx: I tested with this code and I think the exception will be re-throwed.

```php
<?php

function throwException() {
    throw new Exception('test');
}

try {
    print("Outer Try\n");
    try {
        print("Try\n");
        throwException();
    } finally {
        print("Finally\n");
    }
} catch (\Throwable $th) {
    printf("Outer Catch: %s\n", $th->getMessage());
} finally {
    print("Outer Finally\n");
}

```

Output:

```php
Outer Try
Try
Finally
Outer Catch: test
Outer Finally
```

tienvx (approved):

PR #289, opened 2023-02-13
feat: update phpstan/phpstan analyze tool

WaylandAce:
* Update phpstan/phpstan to actual version to resolve tons of deprecation warnings.
* specify phpunit versions to avoid install phpunit 10, since they removed some stuff and pact-php not working with it.
* Fix tests namespace to match psr-4 autoloader and avoid additional warnings.


cfmack:
Awesome!  Thank you.

PR #292, opened 2023-03-02
Use Rust FFI #3.2: Default port zero

tienvx:


YOU54F:
Yep totally in agreement that defaulting to port 0 and therefore allowing the rust core to try and find an available port, would be the preferred behaviour for all users, with them otherwise having the option to select their own

YOU54F (approved):
I would in agreement that a default port zero is a very sensible default, the core try and find an available port

YOU54F:
merged in #323 

https://github.com/pact-foundation/pact-php/commit/2706eff5680078f4c43dfb28dd6ee6b92d44ec48

tienvx:

PR #293, opened 2023-03-02
Use Rust FFI #3.3: Default write mode merge

tienvx:


tienvx:
Close because it conflict with https://github.com/pact-foundation/pact-php/pull/302. Default write mode merge is also included in that PR

PR #295, opened 2023-03-06
Use Rust FFI #6.1: Remove stub server health check

tienvx:
Depend on https://github.com/pact-foundation/pact-php/pull/283

This feature is not available in [Pact Stub Server (Rust)](https://github.com/pact-foundation/pact-stub-server)


tienvx:
Code is moved to https://github.com/pact-foundation/pact-php/pull/283

PR #298, opened 2023-03-06
Use Rust FFI #3.4: Make tag optional

tienvx:


src/PhpPact/Consumer/Listener/PactTestListener.php
```diff
@@ -73,16 +73,16 @@ public function endTestSuite(TestSuite $suite): void
                 print 'PACT_BROKER_URI environment variable was not set. Skipping PACT file upload.';
             } elseif (!($consumerVersion = \getenv('PACT_CONSUMER_VERSION'))) {
                 print 'PACT_CONSUMER_VERSION environment variable was not set. Skipping PACT file upload.';
-            } elseif (!($tag = \getenv('PACT_CONSUMER_TAG'))) {
-                print 'PACT_CONSUMER_TAG environment variable was not set. Skipping PACT file upload.';
```
tienvx, line 102: I think tag is not required for `pact-broker publish` command.

src/PhpPact/Consumer/Listener/PactTestListener.php
```diff
@@ -95,7 +95,6 @@ public function endTestSuite(TestSuite $suite): void
                 }
 
                 $broker = new Broker($brokerConfig);
-                $broker->createVersionTag();
```
tienvx: Removed because:
1. Tag will be created by `$broker->publish();` line below if provided.
2. If the the version of consumer didn't exist in Pact Broker before, this line will cause this error:
```
Could not create tag. Version [VERSION_HERE] of [CONSUMER_NAME_HERE] does not exist.
```

tienvx:
Close because it conflict with https://github.com/pact-foundation/pact-php/pull/302

PR #299, opened 2023-03-06
Use Rust FFI #3.5: Support branch

tienvx:
Depend on https://github.com/pact-foundation/pact-php/pull/298


tienvx:
Close because it conflict with https://github.com/pact-foundation/pact-php/pull/302

PR #301, opened 2023-03-13
Run php cs fixer on PHP 8.2

tienvx:
PHP-CS-FIXER released version 3.15 with PHP 8.2 support

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [ '8.0', '8.1' ]
+        php: [ '8.0', '8.1', '8.2' ]
 
     steps:
       - name: 🛑 Cancel Previous Runs
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [ '8.0', '8.1', '8.2' ]
+        php: [ '8.2' ]
 
     steps:
       - name: 🛑 Cancel Previous Runs


sergeyklay:
Я думаю, чтобы ускорить проверку стиля кода и не делать лишней работы, достаточно использовать последнюю версию PHP. Я не могу представить себе ситуацию, когда линтер под PHP 8.2 не найдет того, что найдет линтер, работающий под PHP 8.1.

sergeyklay:
I think to speedup code style checks and not to make redundant work it is enough to use the latest version of PHP. I can't imagine a situation where a linter under PHP 8.2 won't find something that a linter working under PHP 8.1 will find.

tienvx:
Hi @sergeyklay I agree with you. I also has similar question in my previous PR https://github.com/pact-foundation/pact-php/pull/278#discussion_r1048052121

I removed PHP 8.0 and 8.1

YOU54F:
Thanks for helping with the review on this @sergeyklay, @tienvx will greatly appreciate the feedback, it has been a slow cycle in pact-php but we would love to keep it going for our community. We've been lucky to have a few collaborators over the last year but some of the similar issues to pact-python regarding maintainership, and being able to help with reviews and nudge stuff along.

Tien has kindly stepped up to the plate on the back of great work done by @mmoll / @Lewiscowles1986 and others.

sergeyklay:
@tienvx I have experience with PHP, so ping me anytime

Lewiscowles1986:
Not sure why I'm being thanked, but I feel awkward now for not being more active.

https://github.com/pact-foundation/pact-php/actions/runs/4403773326/jobs/7712576073?pr=301 seems to have failed, but I'm betting that a retry will see it green.

Is there perhaps some way to log separately to a file on the entry and exit to each flow?

Where this launches processes still, I am sure that is where the problem lays. I definitely cannot look this weekend (UK Mothers day), but deterministic test-runs to the shared network resource might be the ticket (might). This should not need to guarantee test case ordering, just that the steps in every test case are clean and have no races.

YOU54F:
>Not sure why I'm being thanked, but I feel awkward now for not being more active

Sorry just thanking for past efforts to date, not wanting to make you feel bad or trying to pull you into help more. We are keen to act on some of your prior comments as your feedback was incredibly useful Lewis, so just trying to bring others into the fold and thank others in the process!

PR #300, opened 2023-03-07
Use Rust FFI #9: Support multiple provider states with params interaction

tienvx:
This pull request depend on https://github.com/pact-foundation/pact-php/pull/285, please review it first.

https://github.com/pact-foundation/pact-specification/tree/version-3#allow-multiple-provider-states-with-parameters

PR #302, opened 2023-03-13
Use Rust FFI #10: Remove pact broker

tienvx:
This PR follow suggestion from @mefellows that:
* Don't embed cli (inside Ruby Standalone) into Pact-PHP
* Recommend most users of Pact  to pick your client language (e.g. PHP) and then download the CLI tools to do the CI/CD integration

One of the reason is: The Ruby Standalone has windows long file path and not working on a bunch of OS/Arch combinations. So bundling it into Pact PHP will mean Pact PHP won’t be installable on those same systems

@mefellows will give further explanation if needed.

I think this suggestion make sense because:
* This speed up the tests, because if we register more than 1 listener, they will publish pacts multiple times.
* Sometime we just want to run the tests, but don't want to publish pacts (e.g. local development?)
* Let imagine there are 3 test suites (Unit, Contract, Integration) running in this specific order. If Unit and Contracts tests are passed, the listener publish pacts, but then Integration tests failed. Then I think we published some pacts that we don't really want.

For Github Actions it's quite easy to add a step to publish pacts while still follow other recommendation (like include commit SHA in the version, use branch instead of tag):

```
    - name: Publish pacts
      uses: pactflow/actions/publish-pact-files@auto_detect_version_branch
      env:
        pactfiles: tests/Contract/pacts
        PACT_BROKER_BASE_URL: ${{ secrets.PACT_BROKER_BASE_URL }}
        PACT_BROKER_TOKEN: ${{ secrets.PACT_BROKER_TOKEN }}
```

This PR depend on https://github.com/pact-foundation/pact-php/pull/300. Please review it first.


YOU54F:
Thanks for the explanation and detail @tienvx 

I think it is sensible to split out the pact-broker functionality and for users to use the CLI tools instead, which as you've said can be easily installed and used in CI.

We are looking to improve the ways users can get them on their machine ( scoop / choco / homebrew / snap / flatpak etc)

We have the same consideration in pact-js

Same proposal to move it out

-https://github.com/pact-foundation/pact-js-core/pull/435

In to a standalone CLI in JS

https://github.com/pact-foundation/pact-js-cli

I wonder if enough users warrant having the functionality in their native language (and can support it) then it could become a separate PHP package. Your download tool for composer would make it quite neat (that way we don't lose any of the good work done so far)

and with a reboot of the pact-ruby-standalone  for more arches and latest ruby version there is some value I reckon

https://github.com/pact-foundation/pact-ruby-standalone/pull/100

Plus I would hope that community contributors might relish the thought of being able to hack on a new CLI wrapper in their native language away from the trappings of the larger language it resides in.

Food for thought :)

Thanks dude

tienvx:
> it could become a separate PHP package

@YOU54F thanks for the idea. I will take it into account when this PR is being reviewed.

YOU54F:
merged in #323 

https://github.com/pact-foundation/pact-php/commit/e40184753fe4f7fecd39f93839af738ce588d749

tienvx:

PR #304, opened 2023-04-27
Fix tests

YOU54F:
the pact-broker client doesn't accept uriencoded characters, and requires spaces to be wrapped.

Screenshot 2023-04-27 at 21 43 28

### DIFF
--- a/tests/PhpPact/Standalone/Broker/BrokerTest.php
+++ b/tests/PhpPact/Standalone/Broker/BrokerTest.php
@@ -50,7 +50,7 @@ class BrokerTest extends TestCase
     public function describeVersion(): void
     {
         $config = new BrokerConfig();
-        $config->setPacticipant("Animal Profile Service")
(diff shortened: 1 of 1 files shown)


YOU54F:
I'll work out if this is a regression 

a) in the pact-broker client (did it ever accept uri encoded spaces
b) at PactFlow's end, at this is calling the PactFlow hosted test broker

I do think it would be good to 

a) use a locally spun up broker (its problematic on windows/macos runners, as no docker, so would have to be with ruby - although may it could be setup via an action?)

b) if it is going to use the PF broker, all the data used should be setup by the test,(the "Animal Profile Service" is used in a few of the pact repos, but there isn't an equivalent in here 

(The previous contributor who added the test, probably added it based on an example elsewhere, where it did setup the data)

It is the challenge when a user just wants a single feature, implements in and makes a test with the best intentions.

Hopefully this helps people out proceeding and getting them sweet sweet green builds

PR #306, opened 2023-04-28
broker: Fix broker describe pacticipant argument passing

Greg0:
In reference to https://github.com/pact-foundation/pact-php/pull/304 
Config class end-user should not be aware of correct passing data with them usage in mind (e.g. quotes).
I think `PhpPact\Standalone\Broker\Broker` class should get correct use of config data and pass them to CLI command in right way.

### DIFF
--- a/src/PhpPact/Standalone/Broker/Broker.php
+++ b/src/PhpPact/Standalone/Broker/Broker.php
@@ -36,7 +36,7 @@ class Broker
             \array_merge(
                 [
                     'can-i-deploy',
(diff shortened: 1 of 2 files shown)


YOU54F:
> In reference to #304
> Config class end-user should not be aware of correct passing data with them usage in mind (e.g. quotes).
> I think `PhpPact\Standalone\Broker\Broker` class should get correct use of config data and pass them to CLI command in right way.

appreciate your eyes across that and i agree that makes sense, the user shouldn’t need to know this, and it should be handled by the framework. nice job

PR #307, opened 2023-04-28
Fix/display stderr on publish error

YOU54F:
- centralise logging on non 0 exit code from standalone
- return exit code in exception

includes #288 which makes sense, but worth having in it one central place as they all use the same function 👍🏾 

Thanks @WaylandAce for all the good work


cfmack (approved):

YOU54F:
I can confirm this has worked and helped massively, as I've managed to work out why test was failing in linux on arm64 

and before the error was just lost in the midst of time

PR #315, opened 2023-06-12
MockServerConfig improvements

slt:
* Fix config setters returning a interface which prevents chaining when static analysis tools are in use as the interfaces are only a subset of what the class can do.

* Set defaults similar to MockServerEnvConfig for timeouts as these are typed but not set to a value by default, forcing you to set them to avoid an error.


YOU54F:
complete in https://github.com/pact-foundation/pact-php/pull/322 thanks @slt

PR #316, opened 2023-06-12
Gotta go fast

slt:
Pact consumer tests are supposed to be like unit tests but mine were taking 2 seconds+ each, the majority of which was php sleep()

* Allow SetHealthCheckRetrySec / PACT_MOCK_SERVER_HEALTH_CHECK_RETRY_SEC
  to be a float so it can be set to 0.1 instead of 1.

* Use usleep instead of sleep so that the retry seconds can be a < 1 second.

* Increase the total number of attempts so it still tries for ~10 seconds.

* Remove sleeps in PactMessage which do not appear to be needed.

Results in my environment 

3 pact consumer tests
Before - 00:06.577s
After  - 00:01.603s

32 pact message tests
Before - 00:55.202s
After  - 00:21.900s

### DIFF
--- a/src/PhpPact/Standalone/MockService/MockServer.php
+++ b/src/PhpPact/Standalone/MockService/MockServer.php
@@ -42,10 +42,6 @@ class MockServer
         $processId =  $this->processRunner->run();
 
         $result = $this->verifyHealthCheck();
-        if ($result) {
-            $retrySec = $this->config->getHealthCheckRetrySec();
-            \sleep($retrySec);
-        }
 
         return $processId;
     }
@@ -122,7 +118,7 @@ class MockServer
             try {
                 return $service->healthCheck();
             } catch (ConnectionException $e) {
-                \sleep($retrySec);
+                \usleep(round($retrySec * 1000000));
             }
         } while ($tries <= $maxTries);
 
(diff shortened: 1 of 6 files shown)


mefellows:
This is great, thanks for all of these PR's @slt. I'm not qualified to review the PHP bits, so will wait for a maintainer - but just wanted to pass on my gratitude.

src/PhpPact/Standalone/MockService/MockServer.php
```diff
@@ -122,7 +118,7 @@ private function verifyHealthCheck(): bool
             try {
                 return $service->healthCheck();
             } catch (ConnectionException $e) {
-                \sleep($retrySec);
+                \usleep(round($retrySec * 1000000));
```
YOU54F, line 121: round returns a float and usleep expects an int

https://www.php.net/manual/en/function.round.php
https://www.php.net/manual/en/function.usleep.php

`usleep(intval(round($retrySec * 1000000)));`

YOU54F (changes requested):
the linter is failing, I've noted a comment

YOU54F:
Kicked off a build https://github.com/pact-foundation/pact-php/actions/runs/5556096016

YOU54F:
complete in https://github.com/pact-foundation/pact-php/pull/322

PR #323, opened 2023-07-18
pact_ffi library targeting ffi branch for beta use by users

YOU54F:
This is a continuation of the awesome work done by @tienvx 

Should hopefully contain all of your merge commits from the FFI PR's described in #262 

Additional changes include

- Inclusion of testing of arm64 linux/macos
- Addition of aarch64 detection for linux/macos pact-stub-server
- Added some developer docs
- Updated readme and migration guide to for ffi to target 10.x
- Possibly borked some merge conflicts
  - I assume cores, and health check delays aren't required in the pact ffi version, as they were removed from your code.

Nice work @tienvx - Looking forward to getting this merged into the `ffi` branch for users to test out, and provide feedback. 

I've found some php repos in the wild, so going to see if I can spend some time getting them running, and then update them to use this version, and it will provide a nice delta.

@tienvx Happy if you'd prefer to provide a singular pull request rebased against the `ffi` branch and I can close this off - just had some time and this was helpful to me to review all the pieces in situ

PR #310, opened 2023-05-08
Use Rust FFI #3.6: Add generators and matchers

tienvx:
Replace https://github.com/pact-foundation/pact-php/pull/308

Depend on https://github.com/pact-foundation/pact-php/pull/280, please review it first.

This PR do:
* Add most of matchers in https://docs.pact.io/implementation_guides/rust/pact_ffi/integrationjson, except `eachKey` and `eachValue`
* Add most of generators in https://docs.pact.io/implementation_guides/rust/pact_ffi/integrationjson, except `MockServerURL`
* Some matchers I don't know how to use yet e.g. `contentType` or `statusCode`
* Try to not affect current users by creating separate matcher even though they are the same e.g. `integer` vs `integerV3`
* Some matchers are copied from [pact-js](https://github.com/pact-foundation/pact-js/blob/master/src/v3/matchers.ts) but are modified in the way that I think it's easier **for me** to understand
* Missing matchers are those I don't know how to implement yet: `eachKey` and `eachValue`
* Missing generators are those I don't know how to implement yet: `MockServerURL`
* Since the class `Matcher` getting bigger and bigger, do we need to refactor to make the code cleaner?

PR #303, opened 2023-03-16
Use Rust FFI #11: Allow set content type

tienvx:
This PR refactor the code to support plugin (csv, protobuf)

This PR depend on https://github.com/pact-foundation/pact-php/pull/302, please review it first.

PR #297, opened 2023-03-06
Use Rust FFI #7.1: Stub server random port

tienvx:
Depend on https://github.com/pact-foundation/pact-php/pull/284

PR #326, opened 2023-07-25
Pact PHP - Support for V3 / V4 + Plugins Via FFI

YOU54F:
# Pact PHP - Support for V3 / V4 + Plugins Via FFI

## What is the current status

Pact-PHP does not currently support v3+ features of the rust core, in the mainline branch. 

V3+ features are now available for beta testing in `ffi` branch, in a view to stabilise and release into the mainline.

We would <3 your support

## How can I test this today

Anyone can start testing `pact-php` with `pact_ffi` code by updating their `composer.json`

`"pact-foundation/pact-php": "10.0.0-alpha1"`

- Migration Guide:- https://github.com/pact-foundation/pact-php/blob/ffi/UPGRADE-10.0.md
- [10.0.0-alpha1 release notes](https://github.com/pact-foundation/pact-php/releases/tag/10.0.0-alpha1)

## PR's which made up this release

NB:- Although marked as closed, they have been consolidated and merged into a single branch under #323 

* [X] Consumer #280 
* [X] Message Consumer #281
* [X] Provider #282
* [X] Stub Server #283 
* [X] #294 
* [X] #295 
* [X] #295 
* [X] #296 
* [X] #297 
* [X] #302 
* [X] #303 
* [X] #310 

## How can I submit feedback

All feedback is gratefully received, positive feedback is great, as it allows us to ascertain how well the package is doing for users. Constructive criticism is okay too, but see the next section

- Reply in this pull request
- Raise a new issue, clearly stating you are using the `ffi` branch

## How can I contribute

If you want to help improve Pact-PHP, read the [`DEVELOPING.md`](https://github.com/pact-foundation/pact-php/blob/ffi/DEVELOPING.md) documentation, fork the repo and have a play!


tienvx:
I created `9.x` branch from current `master` branch for future PRs that target standalone ruby.

composer.json
```diff
@@ -18,34 +18,27 @@
         }
     ],
     "require": {
-        "php": "^8.0",
-        "ext-openssl": "*",
```
Lewiscowles1986, line 22: won't this need ext-openssl? Extensions are always a PITA to document

composer.json
```diff
@@ -18,34 +18,27 @@
...
-        "amphp/socket": "^1.1.3",
-        "amphp/sync": "^1.4.0",
-        "amphp/cache": "^1.4.0",
-        "amphp/windows-registry": "v0.3.3",
-        "guzzlehttp/guzzle": "^6.5.8|^7.4.5",
-        "phpunit/phpunit": ">=8.5.23 <10",
-        "tienvx/composer-downloads-plugin": "^1.1.0"
+        "symfony/process": "^5.4|^6.0|^7.0",
+        "guzzlehttp/psr7": "^2.4.5",
+        "tienvx/composer-downloads-plugin": "^1.2.0"
```
Lewiscowles1986, line 27: I know this is a fork of civicrm. I'd feel so much more comfortable if this could be brought under PACT or get a lockfile to ensure the hash of the version is known and audited. This is going to become a runtime package for everyone using PACT going forward.

composer.json
```diff
@@ -55,44 +48,106 @@
...
+            "ProtobufSyncMessageProvider\\Tests\\": "example/protobuf-sync-message/provider/tests",
+            "ProtobufAsyncMessageConsumer\\": "example/protobuf-async-message/consumer/src",
+            "ProtobufAsyncMessageConsumer\\Tests\\": "example/protobuf-async-message/consumer/tests",
+            "ProtobufAsyncMessageProvider\\Tests\\": "example/protobuf-async-message/provider/tests",
+            "StubServerConsumer\\": "example/stub-server/consumer/src",
+            "StubServerConsumer\\Tests\\": "example/stub-server/consumer/tests"
         }
     },
     "scripts": {
-        "start-provider": "php -S localhost:58000 -t example/src/Provider/public/",
```
Lewiscowles1986, line 73: TODO: check this is not referenced and is called out as deprecated. 

```suggestion
        "start-provider": "echo 'removed in 10.x'; exit 1",
```

composer.json
```diff
@@ -55,44 +48,106 @@
...
-                    "{$architecture}": "strtolower(php_uname('m')) === 'arm64' || strtolower(php_uname('m')) === 'aarch64' ? '-arm64' : (strtolower(php_uname('m')) === 'x86' && PHP_OS_FAMILY === 'Windows' ? '-x86' : '-x86_64')",
-                    "{$extension}": "PHP_OS_FAMILY === 'Windows' ? 'zip' : 'tar.gz'"
+                    "{$os}": "PHP_OS === 'Darwin' ? 'osx' : strtolower(PHP_OS_FAMILY)",
+                    "{$architecture}": "in_array(php_uname('m'), ['arm64', 'aarch64']) ? 'aarch64' : 'x86_64'",
+                    "{$extension}": "PHP_OS_FAMILY === 'Windows' ? '.exe' : ''"
                 },
-                "url": "https://github.com/pact-foundation/pact-ruby-standalone/releases/download/v{$version}/pact-{$version}-{$os}{$architecture}.{$extension}",
-                "path": "bin/pact-ruby-standalone"
+                "url": "https://github.com/pact-foundation/pact-stub-server/releases/download/v{$version}/pact-stub-server-{$os}-{$architecture}{$extension}.gz",
+                "path": "bin/pact-stub-server/pact-stub-server{$extension}",
```
Lewiscowles1986, line 137: very clever; I don't see away around having something like this, but... Risk++

This affects BOM as well as clarity, and would be outside of any lockfile.

Lewiscowles1986:
Most of the points I'm calling out here have been called out before, so I don't think we should block this great performance, functionality and stability win.

* I've skipped reviewing all tests and compatibility; PR is where they should have been reviewed and I've done my best there
* Using checks as a guide as well (this is so much more stable than PACT prior)
* Reviewed the patch https://patch-diff.githubusercontent.com/raw/pact-foundation/pact-php/pull/326.patch with both manual review and automated tools looking for anything that could ruin anyones day.

1. Suggestion added for release mentioning 10.x (is this 9.x or 10.x?) adjust as appropriate
2. Comment on moving pieces folks might need a guided hand on

Fantastic Work @tienvx 🎉 

This might take me some time to finish reviewing, but I'm over half-way through in first sitting.

JP-Ellis:
Just wanted to say, absolutely amazing work @tienvx!! And thank you @Lewiscowles1986 for helping with the review and providing feedback.

This is a behemoth of a PR, and thankfully a lot of the work has been reviewed in previous PRs.

Looking forward to this being merged, and Pact PHP v10 being made widely available!

mefellows:
I ~second~ third this. Amazing work Tien - painstakingly breaking it into _who knows how many_ PRs, and then incrementally improving them with more tests, documentation etc.

Seeing it all in one place like this shows the enormity of what has been done.

Lewis - you deserve a lot of credit, I've seen how quickly you've been reviewing and providing feedback on these PRs. 

Tien might be the hero, but you're the Robin to his Batman👏 .

Can't wait to see this in action!

src/PhpPact/Standalone/ProviderVerifier/Verifier.php
```diff
@@ -2,264 +2,198 @@
...
+                $transport->getPath(),
+                $transport->getScheme()
+            );
         }
+    }
 
-        if ($this->config->getFormat() !== null) {
-            $parameters[] = "--format={$this->config->getFormat()}";
-        }
+    private function setFilterInfo(VerifierConfigInterface $config): void
```
silverblaze404, line 69: `setProviderState` and `setFilterInfo` function names got interchanged I think. Might not affect functionality but since I was reading this implementation thought should point this to you :)

tienvx: You are correct. Nice catch. I will create a PR to update it.

composer.json
YOU54F: So this comes up often and it's seems like its been a tough nut to crack over the course of the Pact project.

Pact isn't usable without the shared libs (whether ruby or rust, due to a shared core)

Some client languages need to provide additional binding files, aswell as the shared core.

The shared core is built for different platforms and architectures.

Some languages do not easily provide a mechanism to satisfactorily provide platform/arch specific packages, in a single meta package (Python does for, JavaScript doesn't)

So you end up in a case where you have

- A single package with every platform/arch combo of shared library in
  - easiest for users, everything in one place
  - big fat bundles with wastage, if users never run on certain platforms
  - overkill on package managers that can support platform/arch specific combos
- Multiple packages, one for each platform/arch combo
  - Ideal if package manager can detect
  - Less than ideal if package manager cannot detect, and onus is on the user to select the right package, for the selected platform/arch. This can be difficult if you run a project on multiple platform/archs. Needs lots of documentation
- Really slim package that downloads runtime libraries for specified platform/combo
  - Smallest package size possible
  - Requires external files to be useful (needs the shared core)
  - big issue for users with corp proxies, no external access
  - needs ability for users to provide their own binaries and skip downloading of files
  
The shared libraries can be highly trusted as they are generated from CI and pulled from releases, they can be shasum checked, as sha's as also published. The sha's could be stored in the released client libraries packages such that if an attacker updates the pact ref releases, and the associated sha, the client library package would reject it, as it stores it's own copy for of the shasum
  
I do wonder how much users worry about the size, for me, it is a real concern as I run multiple example projects so have huge numbers of duplication, of binaries/executables that can be made smaller (something else I am looking at)

It may be also useful to consider where practical for the shared library to be installed to a common location such as a users `$HOME/.pact` library, and client libraries looking there. That way we could have an installer that can deal with getting you a known shared library, and the pact client libraries don't have to concern with packaging it, just referencing it by version in a known location.

Surfacing the version used of the client libraries easily, in vanilla client language without requiring ffi func or extensions gives users the best chance of being able to find out the version of the pact client libs they need, so they can install them.

I'd like alot of these concerns really to be taken up with a pact cli tool that can install/audit and verify the shared libs, and a doctor command that can check an env/system's suitability for each client lang

src/PhpPact/Standalone/ProviderVerifier/Verifier.php
tienvx: Fixed by https://github.com/pact-foundation/pact-php/pull/542

composer.json
tienvx: > The shared libraries can be highly trusted as they are generated from CI and pulled from releases, they can be shasum checked, as sha's as also published. The sha's could be stored in the released client libraries packages such that if an attacker updates the pact ref releases, and the associated sha, the client library package would reject it, as it stores it's own copy for of the shasum

I think it's a great idea, and we can add this feature to https://github.com/tienvx/composer-downloads-plugin (or https://github.com/pact-foundation/composer-downloads-plugin if we made it available first)

Lewiscowles1986:
@YOU54F https://github.com/Lewiscowles1986/pact-testing/actions/runs/8291625745/job/22691646516 which is a fork of the first you showed, says this works

- deleted the lockfile to pull from `ffi` branch and observed improved readability of output
- Experimented until landing on `--testdox` style rather than `dots` default

composer.json
Lewiscowles1986: So I don't disagree with anything said here, and I don't think anyone should feel their position is invalid.

Tien, your suggestion looks like it would mitigate a lot of the risk.

What exists works. The continued call-out are improvement suggestions. The more objections the project can overcome, the easier adoption should become. PACT is a heavyweight implementation to existing non-greenfield, although the barrier for greenfield is low.

composer.json
tienvx: Added `ext-openssl` again in https://github.com/pact-foundation/pact-php/pull/541

composer.json
Lewiscowles1986: Nothing references that I have seen, so maybe not needed. I Just appreciate QoL features like this.

Lewiscowles1986 (approved):
LGTM

composer.json
tienvx: Added again in https://github.com/pact-foundation/pact-php/pull/543

composer.json
Lewiscowles1986: @tienvx thanks for continuing to put up with my suggestions etc. Now this is within `pact-foundation` I think the sha checking can be a separate future-thing

tienvx: > is there anyone that can help get a packagist package published for that repo now it's within pact-foundation?

Let me ask in Slack Maintainer channel.

composer.json
tienvx: Replaced in https://github.com/pact-foundation/pact-php/pull/544

composer.json
tienvx: Look like he is very busy. We can use `repositories` feature of Composer to allow installing the package for now  https://github.com/pact-foundation/pact-php/pull/544

PR #327, opened 2023-07-25
chore: Replace atLeastAndMostLike by arrayLike

tienvx:
`atLeastAndMostLike` is my invention, and I don't think it's useful. Instead, I replace it with `arrayLike` (which is also my invention).

`arrayLike` do what `atLeastAndMostLike` do, which is checking both `min` and `max`, but it also:

* Allow any array as example value (so example value doesn't need to be a value appear multiple times)
* Allow min = 0 (in this case example value doesn't need to be empty array)
* Min can be null
* Max can be null
* Min and max can be both null (now it become `$matcher->like()`)
* Min and max are not null (`atLeastAndMostLike` above)

for example:

```php
$user = [
    'gender' => $matcher->regex('male', 'male|female|other'),
    'firstName' => $macher->like('Bettye'),
    'lastName' => $macher->like('Johnston'),
    'birthday' => $matcher->date('yyyy-MM-dd', '2022-11-21')
];
$father = [
    'gender' => 'male',
    'firstName' => 'Deonte',
    'lastName' => 'Zulauf',
    'birthday' =>  '1970-11-21'
];
$mother = [
    'gender' => 'female',
    'firstName' => 'Delpha',
    'lastName' => 'Ryan',
    'birthday' =>  '1986-11-21'
];
$users = $matcher->arrayLike([
    $user,
    $father,
    $mother
], 0)
```

### DIFF
--- a/src/PhpPact/Consumer/Matcher/Matcher.php
+++ b/src/PhpPact/Consumer/Matcher/Matcher.php
@@ -93,23 +93,24 @@ class Matcher
     }
 
     /**
-     * @param mixed $value example of what the expected data would be
-     * @param int   $min   minimum number of objects to verify against
+     * @param array $values example of what the expected data would be
+     * @param int   $min    minimum number of objects to verify against
+     * @param int   $min    maximum number of objects to verify against
      *
      * @return array<string, mixed>
      */
-    public function atLeastAndMostLike(mixed $value, int $min, int $max): array
+    public function arrayLike(array $values, ?int $min = null, ?int $max = null): array
     {
-        if ($min <= 0 || $min > $max) {
+        if (($min !== null && $min <= 0) || ($min !== null && $max !== null && $min > $max)) {
             throw new Exception('Invalid minimum number of elements');
         }
 
         return [
-            'value' => array_fill(0, $min, $value),
+            'value' => $values,
             'pact:matcher:type' => 'type',
-            'min' => $min,
-            'max' => $max,
-        ];
+        ] +
+        ($min === null ? [] : ['min' => $min]) +
+        ($max === null ? [] : ['max' => $max]);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mefellows:
Hey Tien,

> Allow min = 0 (in this case example value doesn't need to be empty array)

This scenario is not allowed, see https://docs.pact.io/faq#why-is-there-no-support-for-specifying-optional-attributes. 

I'd suggest modelling based on the names here, for consistency across languages: https://github.com/pact-foundation/pact-js/blob/master/src/v3/matchers.ts

tienvx:
I updated my example so it make more sense why I use `min = 0` there

> This scenario is not allowed, see https://docs.pact.io/faq#why-is-there-no-support-for-specifying-optional-attributes.

Yes, I agree with it. But it's a pain to implement it. e.g. I will need to split a single test to multiple tests:
* Users with/without spouses
* Users with/without father
* Users with/without mother
* Users with/without children

I tried `min = 0` and it still work, so I need to check is it a bug or not.

> I'd suggest modelling based on the names here, for consistency across languages: https://github.com/pact-foundation/pact-js/blob/master/src/v3/matchers.ts

Let me mark this PR Draft for now, I will continue trying to make `pact-php` more consistent with `pact-js`

YOU54F:
Looks good to me, 

Could we update the matchers docs for this matcher

https://github.com/pact-foundation/pact-php/blob/ffi/README.md?plain=1#L154

example from the pact-js

https://github.com/pact-foundation/pact-js/blob/38a68fb4569a20f54160333498b8b0bc20ec7838/docs/matching.md?plain=1#L218

Note:- probably need to update the all the matchers there but can do that in a sep PR

YOU54F (approved):
Just a suggestion on updating README, otherwise great work

YOU54F:
Just requesting your eyes across this @mefellows

tienvx:
Update `README.md` with `constrainedArrayLike` matcher only

YOU54F:
Thanks @tienvx 

I'll get this merged and then get `10.0.0-alpha2` out for you

mefellows:
Thanks, the constrained version of the function looks OK to me

> Yes, I agree with it. But it's a pain to implement it. e.g. I will need to split a single test to multiple tests:

Yes, it is, but that's kind of the price you pay for those guarantees :)

PR #330, opened 2023-07-29
feat: Use auto-installing plugins feature

tienvx:
No need to set `PACT_PLUGIN_DIR` to custom plugins directory for verifier. Plugins are auto installed to `~/.pact/plugins`


YOU54F:
hmmm, i wonder if we should have the ability to set the pact plug-in dir and for the user to disable auto downloading of plugins in case they feel it's a security concern.

i know there was issues before with the ruby standalone where it would attempt to download for each test but was resolved by your download at install time changes

mefellows:
> hmmm, i wonder if we should have the ability to set the pact plug-in dir and for the user to disable auto downloading of plugins in case they feel it's a security concern.

If you use the pact plugin CLI to install the plugins (either during the build or baked into the runtime image), you can control where the plugin comes from in a safe manner (or even bypass the CLI altogether). I don't think it supports a custom directory though.

> i know there was issues before with the ruby standalone where it would attempt to download for each test but was resolved by your download at install time changes

If the plugin is already present on the machine, then the runtime won't try and install dynamically. Exposing that to client languages makes sense though.

YOU54F:
Danke,I think some of thought pattern for why this might be an issue, for pact-php users is due to this issue

[Install the pact binaries from composer instead of within the code #276
](https://github.com/pact-foundation/pact-php/issues/276)

>If the plugin is already present on the machine, then the runtime won't try and install dynamically. Exposing that to client languages makes sense though.

 I need to check that out in pact-php, I believe that will be the case.



>If you use the pact plugin CLI to install the plugins (either during the build or baked into the runtime image), you can control where the plugin comes from in a safe manner (or even bypass the CLI altogether). I don't think it supports a custom directory though.

Ahhh as in the `pact-plugin-cli` doesn't support an option to unpack to a different dir.

Just as an aside, we did get asked a q on how to get plugins in CI (for pact-js) via [Slack](https://pact-foundation.slack.com/archives/CEQBDD5U4/p1690806986980089?thread_ts=1688563315.039959&cid=CEQBDD5U4)

>Appreciate all the help in getting started. I've got a flow working with our solution!
It's manual at the moment though and I'd like to hook it up to the CI and the broker. I mostly know what I'm doing here as I've done it before but one part I'm unsure on is the plugin itself. For local dev I installed through the cli, but what's the best to do this for a CI solution using github actions?



You can use a bash script for now.

- https://github.com/pact-foundation/pact-js/blob/master/scripts/install-plugins.sh
- https://github.com/pact-foundation/pact-plugins/blob/feat/pact-csv-plugin-linux-aarch64/scripts/install-plugin-cli.sh
- https://github.com/YOU54F/pact-python/blob/feat/ffi_consumer/script/install_plugins.sh

The pact-ruby-standalone also comes with the pact-plugin-cli now installed. (note to myself:- I need to update the docs to reflect this)

- https://github.com/pact-foundation/pact-ruby-standalone

tienvx:
* If I install plugins outside of `~/.pact/plugins` and run `PACT_PLUGIN_DIR=/path/to/plugins composer test`, test cases pass and `~/.pact/plugins` are still empty
* Then I run `composer test` only, test cases still pass, and `~/.pact/plugins` contains downloaded plugins

So I confirm it has ability to disable auto downloading plugins feature.

Tested on:
* https://github.com/tienvx/pact-php-csv
* https://github.com/tienvx/pact-php-protobuf

PR #331, opened 2023-08-21
chore: reorganize examples (create separate directory for each example)

tienvx:
Move examples into sub-directories so we can add more examples later


YOU54F:
Sounds good to me, just a note, there will be some refs to examples in the readme, that might need to be updated (there are also refs to older quite old ref examples, which may be worth removing, or updating to also showcase the new interface)

YOU54F (approved):

tienvx:
Thank you. I forgot this file. I updated some links in README.

PR #332, opened 2023-08-21
Use Rust FFI #12: Binary

tienvx:
Add binary file support with example.

This PR depends on https://github.com/pact-foundation/pact-php/pull/331. Please review it first.

PR #335, opened 2023-08-29
Use Rust FFI #14: Xml example

tienvx:
Add xml example.

This PR depends on https://github.com/pact-foundation/pact-php/pull/332. Please review it first.


example/xml/pacts/xmlConsumer-xmlProvider.json
```diff
@@ -0,0 +1,122 @@
...
+      ],
+      "request": {
+        "headers": {
+          "Accept": "text/xml; charset=UTF8"
+        },
+        "method": "GET",
+        "path": "/movies"
+      },
+      "response": {
+        "body": "<?xml version='1.0'?><movies><movie><title>PHP: Behind the Parser</title><characters><character><name>Ms. Coder</name><actor>Onlivia Actora</actor></character><character><name/><actor/></character></characters><plot>So, this language. It's like, a programming language. Or is it a\nscripting language? All is revealed in this thrilling horror spoof\nof a documentary.</plot><great-lines><line>PHP solves all my web problems</line></great-lines><rating type='thumbs'>7</rating><rating type='stars'>5</rating></movie></movies>",
```
YOU54F, line 21: 😄

example/xml/consumer/tests/Service/HttpClientServiceTest.php
```diff
@@ -0,0 +1,94 @@
+<?php
+
+namespace XmlConsumer\Tests\Service;
+
+use Tienvx\PactPhpXml\Model\Options;
+use Tienvx\PactPhpXml\XmlBuilder;
```
YOU54F, line 6: We will probably want to document that this comes from a seperate-repo that would need to be added, into a readme of the changes.

https://github.com/tienvx/pact-php-xml

ps, what is the advantage of publishing these are separate packages? 

Is it useful for 

- users of the library?
- testing of the library?

or are there other advantages, outside of the possibly maintainence sprawl of having lots of different projects.

It might be that it works for PHP and is what users expect, but equally leads to more places to maintain, and more things to convey to the user, that they need, conditionally, depending on their particular use case

tienvx: The reason I put some code outside of this PR, into separate project `pact-php-xml` because:
* I just want to make this PR small so it easier to review
* The code of that library is highly opinionated
* That library is optional, can easily replace by raw json

What do you think if I move the code back to this PR?

YOU54F: Not sure on a definitive answer, but good question and counter points. I've tagged a few other parties who may be offer to offer their thoughts.

On the fact its optional and highly opinionated, would you see this kind of library either coming under the pact-foundation, or us having a list of them in the repo, so that others can create their own and point to them (many ways to skin a cat and all that!) 

Hey team, just wondering if you can cast your eyes on the above thread. 🙏🏾 

@Greg0 @WaylandAce @Lewiscowles1986 @mmoll @cfmack

Lewiscowles1986: reading now, but feel it would be better for Pact to have this repo under their umbrella with a contributor to review (there might be a lot I'd miss as I dip in and out)

Lewiscowles1986: Regarding the specifics, it's a value-object using builder; it's a strange mix. Why this vs templating? Unless I'm mistaken (in which case I apologise); but it looks like a special string builder. XML is not strictly speaking a string.

example/xml/consumer/tests/Service/HttpClientServiceTest.php
```diff
@@ -0,0 +1,94 @@
...
+                            of a documentary.
+                            EOF
+                        )
+                    ->end()
+                    ->start('great-lines')
+                        ->eachLike('line')
+                            ->contentLike('PHP solves all my web problems')
+                        ->end()
+                    ->end()
+                    ->start('rating', ['type' => 'thumbs'])
```
Lewiscowles1986, line 55: @YOU54F It's your library, what do you think of end-users needing to do this?

tienvx: Do you mean this?

```
->start('rating')
    ->attribute('type', 'thumbs')
    ->contentLike(7)
->end()
```

YOU54F: It looks ok to me , and its a decent start. I'm not familiar enough with PHP to know if its feel natural enough for users expected experience, but I can run the tests and get a pact expressing my xml content, with matchers. That capability versus not having it, makes me think whatever works, works for now, and we can iterate in the future.

Working with XML is never going to be nice, especially if you have huge objects, and I think for a better experience overall, if you are using an XML format, such as SOAP, building out a pact-plugin to deal with that might make a better experience.

Do you have reservations with the approach?

As an aside, I know that in Java world, users have difficulty generating pact interaction definitions in the standard library and there are a few DSL builders provided by community, so what is being proposed here isn't alien.

I'm definitely curious if users want slimmed down core package, with the ability to bolt in what they need, when they need so they can mix and match, or is it better just to have everything in one box. I know there will be people on both sides on the fence, and I often flip flop (I might take an OSS project, fork it and strip out extraneous things I don't need, for my purpose - niche case but you know :D)

example/xml/consumer/tests/Service/HttpClientServiceTest.php
YOU54F: I believe it to be a special string builder, as we need to store the XML in the pact json file, so its stored in a string (along with any matchers that have been encoded as part of the test)


This is unpacked at the provider side, and reads the content type of application/xml and sends the content appropriately to the provider

example/xml/consumer/tests/Service/HttpClientServiceTest.php
Lewiscowles1986: It's totally unnatural in any programming language to need to coordinate start and end. I'd suggest looking at functional options pattern. Basically using a spread operator, with functions which are supplied via spread, receive and modify an internal state object (can be idempotent read-only objects if needs be).

That would mean each function would automatically start and end tags, so it could be implicit, but it's still quite heavy-weight compared to supplying XML, because it's such an open format.

example/xml/consumer/tests/Service/HttpClientServiceTest.php
Lewiscowles1986: `json_encode($iContainXML)` should handle escaping etc though right?

I guess we need to know if pact needs UTF-8 escaped, but XML does not and JSON spec does not, so I'm guessing not. 😄

Lewiscowles1986:
Is there a concrete example of the XML this produces, or the "string with XML" this produces? I'd very much like to compare it to `json_encode` on the same string.

tienvx:
> Is there a concrete example of the XML this produces, or the "string with XML" this produces?

Pact's Rust core (FFI) required the body in JSON format (not XML). Here is the json string (json encoded from builder's array):

```json
        {
            "version": "1.0",
            "charset": "UTF-8",
            "root": {
                "name": "movies",
                "children": [
                    {
                        "pact:matcher:type": "type",
                        "value": {
                            "name": "movie",
                            "children": [
                                {
                                    "name": "title",
                                    "children": [
                                        {
                                            "content": "PHP: Behind the Parser",
                                            "matcher": {
                                                "pact:matcher:type": "type"
                                            }
                                        }
                                    ],
                                    "attributes": {}
                                },
                                {
                                    "name": "characters",
                                    "children": [
                                        {
                                            "pact:matcher:type": "type",
                                            "value": {
                                                "name": "character",
                                                "children": [
                                                    {
                                                        "name": "name",
                                                        "children": [
                                                            {
                                                                "content": "Ms. Coder",
                                                                "matcher": {
                                                                    "pact:matcher:type": "type"
                                                                }
                                                            }
                                                        ],
                                                        "attributes": {}
                                                    },
                                                    {
                                                        "name": "actor",
                                                        "children": [
                                                            {
                                                                "content": "Onlivia Actora",
                                                                "matcher": {
                                                                    "pact:matcher:type": "type"
                                                                }
                                                            }
                                                        ],
                                                        "attributes": {}
                                                    }
                                                ],
                                                "attributes": {}
                                            },
                                            "examples": 2
                                        }
                                    ],
                                    "attributes": {}
                                },
                                {
                                    "name": "plot",
                                    "children": [
                                        {
                                            "content": "So, this language. It's like, a programming language. Or is it a\\nscripting language? All is revealed in this thrilling horror spoof\\nof a documentary.",
                                            "matcher": {
                                                "pact:matcher:type": "type"
                                            }
                                        }
                                    ],
                                    "attributes": {}
                                },
                                {
                                    "name": "great-lines",
                                    "children": [
                                        {
                                            "pact:matcher:type": "type",
                                            "value": {
                                                "name": "line",
                                                "children": [
                                                    {
                                                        "content": "PHP solves all my web problems",
                                                        "matcher": {
                                                            "pact:matcher:type": "type"
                                                        }
                                                    }
                                                ],
                                                "attributes": {}
                                            },
                                            "examples": 1
                                        }
                                    ],
                                    "attributes": {}
                                },
                                {
                                    "name": "rating",
                                    "children": [
                                        {
                                            "content": 7,
                                            "matcher": {
                                                "pact:matcher:type": "type"
                                            }
                                        }
                                    ],
                                    "attributes": {
                                        "type": "thumbs"
                                    }
                                },
                                {
                                    "name": "rating",
                                    "children": [
                                        {
                                            "content": 5,
                                            "matcher": {
                                                "pact:matcher:type": "type"
                                            }
                                        }
                                    ],
                                    "attributes": {
                                        "type": "stars"
                                    }
                                }
                            ],
                            "attributes": {}
                        },
                        "examples": 1
                    }
                ],
                "attributes": {}
            }
        }
```

> I'd very much like to compare it to `json_encode` on the same string.

I'm not sure I understand this one.

example/xml/consumer/tests/Service/HttpClientServiceTest.php
tienvx: Thanks for the suggestion I implemented functional options pattern. Please review again.

This is my first time using that pattern. The code look a bit longer for end-users to write.

example/xml/consumer/tests/Service/HttpClientServiceTest.php
tienvx: I updated the builder to use the Functional Options pattern.

tienvx: - About UTF-8 encoding:
  - According to this [safety note](https://github.com/pact-foundation/pact-reference/blob/master/rust/pact_ffi/src/mock_server/handles.rs#L1344), I think yes, pact (in this case `pactffi_with_body`) need UTF-8 encoding.
  - According to this [doc](https://www.php.net/manual/en/function.json-encode.php#refsect1-function.json-encode-parameters) I think `json_encode` also need the value to be UTF-8 encoded.
- About UTF-8 encoding vs escaping: I'm not good at this. I don't think I handle these things at all in the code. I think we need to add special test cases and then update the code to handle these cases.

Lewiscowles1986:
@tienvx the thing you are building the XML example for. I'd like to see what PHP's native `json_encode` does with the XML generated.

Lewiscowles1986 (approved):
Now it doesn't need coordination of `->start()` and `->end()` calls I think this is a lot easier to work with. The extra package pinned to dev main will need to have something happen before it can be released as that is an attack vector; but I see no problem having this go into `main`

Lewiscowles1986:
@tienvx it's one line you made into another PR; the rest is just refactoring examples to be DRY (they don't have to be, they are separate examples). Does that one line have to be in another MR, or can it just be made here?

Lewiscowles1986:
I've approved the other MR, please rebase this from that branch once #340 is merged, providing this meets the PACT foundation approval as well.

YOU54F:
hmm, I rebased and merged this but this didn't close, maybe because I didn't push the rebase back to the this/original branch?

Following gh instructions in the merge command section above, albeit with the additional rebase

```console
git checkout -b tienvx-xml ffi
git pull git@github.com:tienvx/pact-php.git xml
git rebase -i ffi
<rebase things>
git checkout ffi
git merge --no-ff tienvx-xml
git push origin ffi
```

Lewiscowles1986:
So... Although this passes tests, the pinning to `dev-main` looks like it could lead to issues down the line
`"tienvx/pact-php-xml": "dev-main",`
@tienvx @YOU54F any chance that repo can be transferred into PACT control, under foundation heading, OR use a released tagged version that it is pinned to?

tienvx:
> hmm, I rebased and merged this but this didn't close, maybe because I didn't push the rebase back to the this/original branch?

I think so. But if we push back to the branch of this PR, this PR will become empty. I think we just need to close this PR.

> any chance that repo can be transferred into PACT control, under foundation heading

I'm happy to transfer, but I think this require more work to do, and I didn't do it before:
* Agree about license? That library is currently MIT
* Update package name, from `tienvx/pact-php-xml` to `pact-foundation/pact-php-xml`
* I don't think rename package in packagist is possible, so we need to publish another package named `pact-foundation/pact-php-xml`
* Tag first version `1.0.0`, but please review the code first if are there anything to change.

> OR use a released tagged version that it is pinned to?

Please review that library, are there anything I need to update? then I will tag a version, probably `1.0.0`

YOU54F:
>the pinning to dev-main looks like it could lead to issues down the line
"tienvx/pact-php-xml": "dev-main",
@tienvx @YOU54F any chance that repo can be transferred into PACT control, under foundation heading, OR use a released tagged version that it is pinned to?

agreed, we just hadn't agreed on a path forward atm but thanks for raising again I using a released tagged version that can be pinned to is sensible. It is sufficient to be from a git repo tag, rather than being published to packagist, especially whilst in our alpha phase?

>Agree about license? That library is currently MIT

Licenses across the foundation are a mix of either MIT, or Apache 2.0, and I think everyone is generally comfortable with that. 


related comment in maintainers guide draft https://github.com/pact-foundation/docs.pact.io/pull/270#discussion_r1241329836

> Update package name, from tienvx/pact-php-xml to pact-foundation/pact-php-xml?
>I don't think rename package in packagist is possible, so we need to publish another package named pact-foundation/pact-php-xml

Yes I believe you are correct, that you cannot rename. That is why I believe the original pact-php packages are under mattermacks handle. I am okay with this being outside of the pact foundation and transferred in when you are ready Tien. 

It's not a core piece piece of the project (not all users want/need the xml func) and it is added as a dev dep.

Up to you! If it gets transferred in, you will have admin rights anyway, as its your own repository. We've not taken admin rights away from users who have transferred their repos in, as they are ultimately the original owners and we are but caretakers!

> Tag first version 1.0.0, but please review the code first if are there anything to change.

I would probably start with a <1.x version, until you get some real-world feedback, as users can take a v1.x library as production ready, but that is a matter of personal choice.



Given the below thoughts, it is probably worth tracking this is a separate issue, so we can close this PR down and avoid ancillary noise

>>hmm, I rebased and merged this but this didn't close, maybe because I didn't push the rebase back to the this/original branch?

>I think so. But if we push back to the branch of this PR, this PR will become empty. I think we just need to close this PR.

tienvx:
I release version 0.1.0 of `pact-php-xml` and update it via this PR https://github.com/pact-foundation/pact-php/pull/347

Closed as commits are rebased to `ffi` branch.

PR #339, opened 2023-09-22
Print provider process's log

tienvx:
* Move provider process code to helper class to re-use, reduce duplicated code
* Print out provider process' log to help debugging


JP-Ellis (approved):
👍

YOU54F:
Thanks Tien. I believe the test failures are unrelated here, as they have identified in the master branch as well (and were green a month ago)

tienvx:
Yes, it's not related. This PR only show that error out so we can debug.

PR #346, opened 2023-10-08
Use Rust FFI #15: Remove StubServerHttpService

tienvx:
When stub server is started, it may serve various requests/responses:
* method: GET, POST, PUT, PATCH, DELETE
* any combination of: path, query, headers
* body: json, xml, csv, protobuf, binary, multipart...

I remove `StubServerHttpService` because: It doesn't provide that flexibility. All it does is send a `GET /endpoint` request with a single header `Content-Type: application/json`, and expect response to be json string. Why not let user install the package (e.g. guzzle) and send the request however they want?

I also remove `GuzzleClient` because it doesn't do anything special. All it does is proxy calls to the actual guzzle client.

We can remove those code so we don't have to maintain them.


YOU54F:
make sense it if increases end user flexibility and reduces the code we need to maintain, that is of limited value.

Just as note, you can add the following in your commit message to denote a breaking change.

>BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with [MAJOR](http://semver.org/#summary) in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type.

YOU54F (approved):

PR #349, opened 2023-10-23
chore: Add matchers example

tienvx:
* This example demonstrate how to use matchers, and how they behave in consumer and provider tests
* This act as test cases for future changes to matchers
* Example for query & header matchers will be added later (in other PR), after waiting for response from these threads:
  * https://pact-foundation.slack.com/archives/C02BXLDJ7JR/p1698024661692909
  * https://pact-foundation.slack.com/archives/C02BXLDJ7JR/p1698024666918839
* Example for generators will be added later (in other PR)

PR #350, opened 2023-10-27
Allow like null

tienvx:
The condition `$value !== null` probably related to Ruby core. I tested with `null` value and it works fine. So I remove this condition.

Depend on https://github.com/pact-foundation/pact-php/pull/349, please review it first


YOU54F (approved):

tienvx:
I will merge this PR now since it has been approved.

PR #353, opened 2023-10-28
feat: allow multi values for regex matcher

tienvx:
This feature allow to use `regex` matcher to match multiple values for query parameter and header. See https://github.com/pact-foundation/pact-reference/issues/332

This feature is mostly for **compatibility suite**, in [v2/http_consumer](https://github.com/pact-foundation/pact-compatibility-suite/blob/main/features/V2/http_consumer.feature) feature, with these scenarios:
* Supports matchers for repeated request query parameters (positive case)
* Supports matchers for repeated request query parameters (negative case)

### DIFF
--- a/example/matchers/consumer/src/Service/HttpClientService.php
+++ b/example/matchers/consumer/src/Service/HttpClientService.php
@@ -20,7 +20,7 @@ class HttpClientService
     {
         $response = $this->httpClient->get("{$this->baseUri}/matchers", [
             'headers' => ['Accept' => 'application/json'],
-            'query' => 'ignore=statusCode&pages=2&pages=3',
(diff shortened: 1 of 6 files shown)


tienvx:
Marked as Draft, because I need to test with compatibility suite first. PHP has 2 different behaviors when it come to the multiple values for query parameter.

tienvx:
Marked as ready. I tested with compatibility suite and it works fine.

YOU54F (approved):

YOU54F:
Looks like we've ran out of credits on cirrus-ci post changes to their pricing models

https://cirrus-ci.org/blog/2023/07/17/limiting-free-usage-of-cirrus-ci/

so we can ignore the failing cirrus-ci tasks for now

PR #354, opened 2023-10-28
chore: Add status code matcher example

tienvx:
This matcher can't be used to match the response code unfortunately, because currently `pactffi_response_status` only allow integer status.

But it can be useful to match a value in body (like this example), query parameter or header.

### DIFF
--- a/example/matchers/consumer/src/Service/HttpClientService.php
+++ b/example/matchers/consumer/src/Service/HttpClientService.php
@@ -3,6 +3,7 @@
 namespace MatchersConsumer\Service;
 
 use GuzzleHttp\Client;
+use Psr\Http\Message\ResponseInterface;
 
 class HttpClientService
 {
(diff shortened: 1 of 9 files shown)


mefellows:
`pactffi_response_status` is only designed for use on the status code of the response, it can't be used elsewhere. 

There is a status code matcher in the core, but it's not yet available via the "integration JSON" (i.e. via FFI)

mefellows:
Actually, it does look like that matcher is available now: https://github.com/pact-foundation/pact-reference/blob/7a1d6b374d1a1b68a50d9cdab62a1ebd554ec70a/rust/pact_ffi/IntegrationJson.md#L1

tienvx:
@mefellows Yes, it's the `statusCode` matcher mentioned in that document.

mefellows:
ha interesting. Are you sure that matcher works on items that aren't the response code? Does it just look at numbers in general?

tienvx:
I tested again, and it doesn't work as expected (any number e.g. 204 still valid for status code serverError).

So this matcher is useless now. Let me update this PR, remove this matcher completely

mefellows:
Ah, I think that might have been why that matcher doesn't work - you can't pass it in to `pactffi_response_status`. I guess we'd need a variant of that method that accepts the matcher?

tienvx:
I checked again, there are several scenarios about that matcher in compatibility suite's  [v4/matching_rule](https://github.com/pact-foundation/pact-compatibility-suite/blob/main/features/V4/matching_rules.feature#L6)

So instead of remove that matcher, I think creating a `pactffi_response_status_v2` like your suggestion is better. I will make this PR as Draft for now.

tienvx:
`pactffi_response_status_v2` is implemented in https://github.com/pact-foundation/pact-reference/pull/346.

This PR is green when that PR is merged and new version is released

tienvx:

PR #357, opened 2023-10-28
feat: Add eachKey matcher

tienvx:
This is a specification v4's matcher

### DIFF
--- a/src/PhpPact/Consumer/Matcher/Matcher.php
+++ b/src/PhpPact/Consumer/Matcher/Matcher.php
@@ -660,4 +660,21 @@ class Matcher
             'pact:matcher:type' => 'contentType',
         ];
     }
+
+    /**
+     * Allows defining matching rules to apply to the keys in a map
+     *
+     * @param array<string, mixed> $values
+     * @param array<mixed>         $rules
+     *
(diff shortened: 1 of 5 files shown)


src/PhpPact/Consumer/Matcher/Matcher.php
```diff
@@ -660,4 +660,21 @@ public function contentType(string $contentType): array
...
+     * @param array<mixed>         $rules
+     *
+     * @return array<string, mixed>
+     */
+    public function eachKey(array $values, array $rules): array
+    {
+        return [
+            'rules' =>             $rules,
+            'value'             => $values,
+            'pact:matcher:type' => 'eachKey',
```
JP-Ellis, line 677: Nitpick: The `=>` aren't aligned. Up to you whether you want to fix this before merging or not, as it is purely stylistic 👍

JP-Ellis (approved):

src/PhpPact/Consumer/Matcher/Matcher.php
tienvx: I fixed it. Thanks for pointing it out.

PR #361, opened 2023-11-17
feat: Support inheritance for body

tienvx:
This feature is useful for implementing compatibility suite.

### DIFF
--- a/src/PhpPact/Consumer/Registry/Interaction/Body/AbstractBodyRegistry.php
+++ b/src/PhpPact/Consumer/Registry/Interaction/Body/AbstractBodyRegistry.php
@@ -23,10 +23,10 @@ abstract class AbstractBodyRegistry implements BodyRegistryInterface
 
     public function withBody(Text|Binary|Multipart $body): void
     {
-        $success = match ($body::class) {
-            Binary::class => $this->client->call('pactffi_with_binary_file', $this->interactionRegistry->getId(), $this->getPart(), $body->getContentType(), $body->getContents()->getValue(), $body->getContents()->getSize()),
-            Text::class => $this->client->call('pactffi_with_body', $this->interactionRegistry->getId(), $this->getPart(), $body->getContentType(), $body->getContents()),
-            Multipart::class => array_reduce(
+        $success = match (true) {
+            $body instanceof Binary => $this->client->call('pactffi_with_binary_file', $this->interactionRegistry->getId(), $this->getPart(), $body->getContentType(), $body->getContents()->getValue(), $body->getContents()->getSize()),
+            $body instanceof Text => $this->client->call('pactffi_with_body', $this->interactionRegistry->getId(), $this->getPart(), $body->getContentType(), $body->getContents()),
+            $body instanceof Multipart => array_reduce(
                 $body->getParts(),
                 function (bool $success, Part $part) use ($body) {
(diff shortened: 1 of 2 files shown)


src/PhpPact/Consumer/Registry/Interaction/Body/AbstractBodyRegistry.php
```diff
@@ -23,10 +23,10 @@ public function __construct(
 
     public function withBody(Text|Binary|Multipart $body): void
     {
-        $success = match ($body::class) {
-            Binary::class => $this->client->call('pactffi_with_binary_file', $this->interactionRegistry->getId(), $this->getPart(), $body->getContentType(), $body->getContents()->getValue(), $body->getContents()->getSize()),
-            Text::class => $this->client->call('pactffi_with_body', $this->interactionRegistry->getId(), $this->getPart(), $body->getContentType(), $body->getContents()),
-            Multipart::class => array_reduce(
+        $success = match (true) {
```
JP-Ellis, line 26: I'm no PHP expert, but what's the point of having a `match (true)`? I understand this to be similar to an `if (true) {...}` which seems redundant. (Of course, I may be quite wrong here).

src/PhpPact/Consumer/Registry/Interaction/Body/AbstractBodyRegistry.php
```diff
@@ -38,7 +38,6 @@ function (bool $success, Part $part) use ($body) {
                 },
                 true
             ),
-            default => false,
```
JP-Ellis, line 41: With the removal of a default, what happens if `$body` is _not_ an instance of either of the classes being tested?

JP-Ellis (commented):
I suspect it is my lack of understanding of PHP, but at least to me, the use of a `match (true)` is confusing.

Would it not be clearer to have:

```php
if ($body instanceof Bytes) { ... }
else if ($body instance of Text) { ... }
// ...
else {
  throw new InteractionBodyNotAddedException();
}
```

Then again, if the `match (true)` pattern is common for PHP, then I'm happy to approve it.

src/PhpPact/Consumer/Registry/Interaction/Body/AbstractBodyRegistry.php
tienvx: It can't be. The type is defined for `$body` already make sure that it must be instance of those classes:

```
public function withBody(Text|Binary|Multipart $body): void
```

This removal is a suggestion by the `PHPStan` tool. If I don't remove it, this PR will failed.

src/PhpPact/Consumer/Registry/Interaction/Body/AbstractBodyRegistry.php
tienvx:
```
$value = match ($x) {
    $y => $z,
};
```

means when `$x` strictly equals to `$y`, value of `$z` is return.

```
$value = match (true) {
    $y => $z,
};
```

is the same, when `$y` is a boolean, and its value is `true`, then `$z` is return.

So it's not the same as `if (true)`, but it's the same as `if ($y === true)`

tienvx:
There are 3 ways of doing this:

- if/elseif/else
- switch (true)
- match (true)

The reason is very simple: I personally prefer `match (true)` because it's new feature from PHP 8, and it's shorter than `switch (true)`. I personally don't like using `if/elseif/else` in this case.

JP-Ellis (approved):

JP-Ellis:
Thanks for explaining this! Indeed it was my misunderstanding of PHP.

PR #362, opened 2023-11-21
feat: Allow capture verifier output

tienvx:
This change is for implementing compatibility suite

### DIFF
--- a/src/PhpPact/Standalone/ProviderVerifier/Model/VerifierLoggerInterface.php
+++ b/src/PhpPact/Standalone/ProviderVerifier/Model/VerifierLoggerInterface.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace PhpPact\Standalone\ProviderVerifier\Model;
+
+interface VerifierLoggerInterface
+{
+    public function log(string $output): void;
+}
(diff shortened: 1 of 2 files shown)


src/PhpPact/Standalone/ProviderVerifier/Model/VerifierLoggerInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace PhpPact\Standalone\ProviderVerifier\Model;
+
+interface VerifierLoggerInterface
```
tienvx, line 5: Not sure this is a good name

JP-Ellis: I think it's best to follow whatever convention this project follows, and then if there is no strong convention here, whatever PHP tends to follow.

I can see below that there is a `VerifierConfigInterface` and therefore `VerifierLoggerInterface` seems quite reasonable.

JP-Ellis (approved):

src/PhpPact/Standalone/ProviderVerifier/Model/VerifierLoggerInterface.php
tienvx: Sorry, I should be more clear about this.

I don't worry about the `XxxInterface` naming convention for now as it's quite common, but the name `VerifierLogger` itself.

I wonder if there are any better name. I did consider `VerifierOutputAwareInterface` but it seems quite long, and and don't know what to name the object that is instance of that interface (i.e. rename `$logger` to what)

JP-Ellis: Well, given that this interface has a single function called `log`, I think that `VerifierLoggerInterface` is quite a good name. The `VerifierOutputAwareInterface`  on the other hand is a bit less clear. So happy for you to go ahead :)

tienvx: Thank you :pray:

PR #363, opened 2023-11-21
feat: Allow capture mock server mismatches

tienvx:


src/PhpPact/Consumer/Service/MockServer.php
```diff
@@ -37,17 +40,19 @@ public function start(): void
...
+                $this->logger->log($this->getMismatches());
             }
+
+            return $matched;
         } finally {
             $this->cleanUp();
         }
-
-        return $matched;
     }
```
JP-Ellis, line 56: It would appear to me that this function may not always return. I see only one `return` statement within the `try { ... }` part.

Coming from Python, the body of a `try` statement might raise an exception before it reaches the end, and therefore there should be another `return` statement in case an error is raised.

Please correct me if I'm wrong though :)

JP-Ellis (approved):
Just one comment (happy to clarify if it doesn't make sense). Otherwise, looks good to me.

src/PhpPact/Consumer/Service/MockServer.php
tienvx: > Coming from Python, the body of a `try` statement might raise an exception before it reaches the end,

It's the same here in PHP.

> there should be another `return` statement in case an error is raised.

In PHP, when an exception is thrown, we can catch it, then do something with it (then optionally rethrow it), or ignore it.

If I catch it, the code will look like this:

```
 catch (Throwable $throwable) {
            if ($this->logger) {
                $this->logger->log($throwable->getMessage());
            }

            return false;
}
```

But in this case I choose to ignore it, and the exception will be popup to higher level, and if the exception is still not catched, an "Uncaught Exception ..." message will be displayed to user.

Not sure I did the right thing there. What do you think?

tienvx: All I want there is: clean up mock server no matter what there is an exception or not. If there is an exception, user has to deal with it

JP-Ellis: Sounds good to me!

As to whether the error should be caught or not, that would depend I think on the kind of error and whether it might be expected or not as part of the `verify()` call. I'll let you decide what you think is best :)

tienvx: Thank you. I will wait for future feedback to decide better.

PR #365, opened 2023-11-29
chore: Add generators example

tienvx:
There is no example for `MockServerURL`  yet, because the `url` matcher is not available in pact-php (yet)


tienvx:
Self reviewed. I will bypass the review process for this PR

PR #366, opened 2023-11-29
chore: Clean up and use matchers/generators for message example

tienvx:
`ProcessText` and `ProcessSong` are basically the same test. I combine them, then add matcher + generator into the example.


tienvx:

PR #367, opened 2023-11-29
feat: Support matchers and generators for message metadata

tienvx:


YOU54F (commented):
nice, great work on adding this into the core

PR #368, opened 2023-11-29
refactor!: Lazy load binary file

tienvx:
* This is a breaking change, but only inside `10.0.0-alpha.x` versions
* Change c-data from managed to unmanaged

### DIFF
--- a/src/PhpPact/FFI/Model/StringData.php
+++ b/src/PhpPact/FFI/Model/StringData.php
@@ -27,7 +27,7 @@ class StringData
     {
         $length = \strlen($value);
         $size   = $length + ($nullTerminated ? 1 : 0);
-        $cData  = FFI::new("uint8_t[{$size}]");
+        $cData  = FFI::new("uint8_t[{$size}]", false);
         FFI::memcpy($cData, $value, $length);
 
         return new self($cData, $size);
--- a/example/binary/consumer/tests/Service/HttpClientServiceTest.php
+++ b/example/binary/consumer/tests/Service/HttpClientServiceTest.php
@@ -14,7 +14,7 @@ class HttpClientServiceTest extends TestCase
 {
     public function testGetImageContent()
     {
-        $imageContent = file_get_contents(__DIR__ . '/../_resource/image.jpg');
+        $path = __DIR__ . '/../_resource/image.jpg';
 
         $request = new ConsumerRequest();
         $request
@@ -26,7 +26,7 @@ class HttpClientServiceTest extends TestCase
         $response
             ->setStatus(200)
             ->addHeader('Content-Type', 'image/jpeg')
-            ->setBody(new Binary($imageContent, in_array(php_uname('m'), ['AMD64', 'arm64', 'aarch64']) ? 'application/octet-stream' : 'image/jpeg'));
+            ->setBody(new Binary($path, in_array(php_uname('m'), ['AMD64', 'arm64', 'aarch64']) ? 'application/octet-stream' : 'image/jpeg'));
 
         $config = new MockServerConfig();
         $config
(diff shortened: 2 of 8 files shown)


src/PhpPact/FFI/Model/StringData.php
```diff
@@ -27,7 +27,7 @@ public static function createFrom(string $value, bool $nullTerminated = true): s
     {
         $length = \strlen($value);
         $size   = $length + ($nullTerminated ? 1 : 0);
-        $cData  = FFI::new("uint8_t[{$size}]");
+        $cData  = FFI::new("uint8_t[{$size}]", false);
```
tienvx, line 30: This is IMPORTANT change.

This change fix this error:

```
There was 1 failure:

1) BinaryConsumer\Tests\Service\HttpClientServiceTest::testGetImageContent
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-Binary String: 0xffd8ffe000104a46494600010100000100010000ffdb004300030202020202030202020303030304060404040404080606050609080a0a090809090a0c0f0c0a0b0e0b09090d110d0e0f101011100a0c12131210130f101010ffc0000b08000f000f01011100ffc400160001010100000000000000000000000000020304ffc40020100002020300020301000000000000000001020304051112132100223151ffda0008010100003f004d95c94d5ba4b4b3bc1bbefe0aea2348d8725ca7d46b9e4f93d01c8da1fc19223767c98c74b2d5ab61600fe72d374818edf8643f6279876da5247e8035d3b06be3459af83ad12cd5e18dc844083c328e39573e97454a8dc6c3909b0c57665153ab1dc7c45382586f63d64890b36d53d412687b3d068e6473b20f4c7f847cffd9
+Binary String: 0x01000000160000000000000000000000ef000000000000002f2a2a0a20202020202a2057686574686572207468652055524920686173207468652064656661756c7420706f7274206f66207468652063757272656e7420736368656d652e0a20202020202a0a20202020202a20605073725c487474705c4d6573736167655c557269496e746572666163653a3a676574506f727460206d61792072657475726e206e756c6c206f7220746865207374616e6461726420706f72742e2054686973206d6574686f642063616e20626520757365640a20202020202a20696e646570656e64656e746c79206f662074686520696d706c656d656e746174696f6e2e0a20202020202a2f005153ab1dc7c45382586f63d64890b36d53d412687b3d068e6473b20f4c7f847cffd9

/home/tien/Projects/pact/pact-php/example/binary/consumer/tests/Service/HttpClientServiceTest.php:51
```

* This change is also needed for compatibility suite. Sooner or later, we need to made this change.
* In theory, we need to call `FFI::free()`
  * But the question is: when, and where?
  * I can't see any memory leak. Probably because of this [explanation](https://vkcom.github.io/kphp/kphp-language/php-extensions/ffi.html#non-owned-unmanaged-memory)

tienvx:

JP-Ellis:
This all looks good for now.

A couple of considerations with the requirements to free string to avoid memory leaks.

1. The Pact FFI is loaded only for the duration of a test and upon completion of the test, the library is unloaded thereby freeing memory anyway. Since it is very unlikely that end-users will have long-running clients leading to memory bloat, I think this is generally a minor issue.
2. Having said the above, it is still a best practice to make sure memory is handled correctly. Could it be worth creating a ticket to track this, and fix this at some point in the future?
3. In Pact Python, I have created a class called an `OwnedString`. When an instance of an `OwnedString` is garbage collected, Python will ensure that the string is appropriately freed. Could something similar be done in PHP?

tienvx:
Thank for taking a look at it. Here is the ticket to track that issue https://github.com/pact-foundation/pact-php/issues/371

PR #369, opened 2023-11-29
chore: Rename StringData to BinaryData

tienvx:
Rename `StringData` to `BinaryData`. This class is only used for binary now. So no need to make it flexible by supporting null-terminated string.


tienvx:

YOU54F (commented):
Makes sense dude, thanks!

PR #373, opened 2023-11-30
chore: Change default datetime format

tienvx:
* Change the default formats. The previous **default** formats I copied them from https://github.com/pact-foundation/pact-specification/tree/version-4#supported-generators. I think the **new** formats make more sense to end-user.
* Doing so also fixed a failed test: on this specific day `2023-11-30`,  the value become `"2023-12-334T17:15:07"`, not `"2023-11-334T17:15:07"` as expected. I'm not sure what is the problem here.

### DIFF
--- a/example/generators/consumer/tests/Service/GeneratorsTest.php
+++ b/example/generators/consumer/tests/Service/GeneratorsTest.php
@@ -44,8 +44,8 @@ class GeneratorsTest extends TestCase
                 'hexadecimal' => $this->matcher->hexadecimal(null),
                 'uuid' => $this->matcher->uuid(null),
                 'date' => $this->matcher->date('yyyy-MM-dd', null),
-                'time' => $this->matcher->time('HH:mm::ss', null),
-                'datetime' => $this->matcher->datetime("YYYY-MM-D'T'HH:mm:ss", null),
+                'time' => $this->matcher->time('HH:mm:ss', null),
+                'datetime' => $this->matcher->datetime("yyyy-MM-dd'T'HH:mm:ss", null),
                 'string' => $this->matcher->string(null),
                 'number' => $this->matcher->number(null),
                 'requestId' => 222,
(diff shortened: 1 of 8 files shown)


tienvx:
* There are 2 "types" of datetime matchers:
  * Specification v2
    * These matchers that are actually regex (specification v2) matcher, with different regex pattern
    * They are:
      * `dateISO8601` use `ISO8601_DATE_FORMAT`  pattern
      * `timeISO8601` use `ISO8601_TIME_FORMAT` pattern
      * `dateTimeISO8601` use `ISO8601_DATETIME_FORMAT` pattern
      * `dateTimeWithMillisISO8601` use `ISO8601_DATETIME_WITH_MILLIS_FORMAT` pattern
      * `timestampRFC3339` use `RFC3339_TIMESTAMP_FORMAT` pattern
    *  I **didn't touch** them
  * Specification v3
    * These matchers that are defined in https://github.com/pact-foundation/pact-specification/tree/version-4#supported-matching-rules:
    * They are:
      * `Timestamp` (or `DateTime`?)
      * `Time`
      * `Date`
    * They need 1 "parameter":
      * `format` e.g. `yyyy-MM-dd HH:mm:ss`
    * The format must be based on the `Java DateTimeFormatter` (this is where I get the info from https://docs.pact.io/implementation_guides/rust/pact_matching#date-and-time-matchers-datetime-date-time)
    * The format is **not related** to PHP at all


About the format problem in those v3 matchers:


* The formats in `Supported matching rules` [section](https://github.com/pact-foundation/pact-specification/tree/version-4#supported-matching-rules) of the specification are:
  * `yyyy-MM-dd HH:ss:mm`
  * `yyyy-MM-dd`
  * `HH:ss:mm`
* The formats in `Supported generators` [section](https://github.com/pact-foundation/pact-specification/tree/version-4#supported-generators) are:
  * `YYYY-mm-DD'T'HH:mm:ss`
  * `yyyy-MM-dd`
  * `HH:mm::ss`

They produce weird value. They are inconsistent. I think some of them are even invalid? I think they are just example values. But I was stupid. I didn't pay attention. I just blindly copied from the `Supported generators` section. And here we are in the situation of the PR's description.



So I suggesting to change the default formats to:
* `yyyy-MM-dd'T'HH:mm:ss` (or even remove the `T` character, to become `yyyy-MM-dd HH:mm:ss` ?)
* `yyyy-MM-dd`
* `HH:mm:ss`

tienvx:
Given the information in the above comment

> it should surely be `2023-12-30T17:15:07`

If you **manually** set the system's date to `30 November 2023`, and then run the test, the month will change from `11` to `12`. Note the date `334` not `30` (but nothing wrong with it, it match the format `DD`, not `dd`).

> What is parsing these odd formats?

I know it's Rust code, but I don't know exactly where it is in the pact-reference code base. I will update when I found it.

> PHP date format is very different from what you have here https://www.php.net/manual/en/datetime.format.php

Yes, it's Java date format, not PHP date format

> this also does not seem to match https://github.com/pact-foundation/docs.pact.io/issues/88#issuecomment-1094986666

Yes, those are regex-based matcher (specification v2), and I didn't touch them

Lewiscowles1986:
Thanks, I really appreciate the consistent efforts you are making; I'm going to recuse myself rather than be unhelpful.

Lewiscowles1986 (approved):

PR #375, opened 2023-12-08
refactor: Add generator classes

tienvx:
Create classes for these [generators](https://github.com/pact-foundation/pact-specification/tree/version-4#supported-generators)

### DIFF
--- a/src/PhpPact/Consumer/Matcher/Exception/InvalidUuidFormatException.php
+++ b/src/PhpPact/Consumer/Matcher/Exception/InvalidUuidFormatException.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace PhpPact\Consumer\Matcher\Exception;
+
+use Exception;
+
+class InvalidUuidFormatException extends Exception
+{
+}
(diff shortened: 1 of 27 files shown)


src/PhpPact/Consumer/Matcher/Exception/InvalidUuidFormatException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace PhpPact\Consumer\Matcher\Exception;
+
+use Exception;
+
+class InvalidUuidFormatException extends Exception
```
Lewiscowles1986, line 7: Interesting, perhaps slightly off-topic observation.

I'd likely have the MatcherException live under `PhpPact\Consumer\Exceptions` and extend something like `PhpPact\Consumer\Exceptions\ConsumerException`, which itself would extend `PhpPact\Exceptions\BaseException`.

Just the naked word `Exception` is a confusing alias for a human reviewer (I'd imagine some static analysis tools may even panic at it's name shadowing a built-in).

tienvx: Added these exceptions in https://github.com/pact-foundation/pact-php/pull/382

PR #379, opened 2023-12-11
refactor: Add RandomHexadecimal generator class

tienvx:
The generator is defined at https://github.com/pact-foundation/pact-specification/tree/version-4#supported-generators

### DIFF
--- a/tests/PhpPact/Consumer/Matcher/Generators/RandomHexadecimalTest.php
+++ b/tests/PhpPact/Consumer/Matcher/Generators/RandomHexadecimalTest.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace PhpPactTest\Consumer\Matcher\Generators;
+
+use PhpPact\Consumer\Matcher\Generators\RandomHexadecimal;
+use PHPUnit\Framework\TestCase;
+
+class RandomHexadecimalTest extends TestCase
+{
+    public function testSerialize(): void
+    {
+        $hexaDecimal = new RandomHexadecimal(8);
+        $this->assertSame(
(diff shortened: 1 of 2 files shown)


tests/PhpPact/Consumer/Matcher/Generators/RandomHexadecimalTest.php
```diff
@@ -0,0 +1,18 @@
...
+class RandomHexadecimalTest extends TestCase
+{
+    public function testSerialize(): void
+    {
+        $hexaDecimal = new RandomHexadecimal(8);
+        $this->assertSame(
+            '{"digits":8,"pact:generator:type":"RandomHexadecimal"}',
+            json_encode($hexaDecimal)
+        );
+    }
```
Lewiscowles1986, line 17: question about tests. Can we have 0-f (it's only 15 cases); can use a table-test if it makes it easier.

tienvx: I don't think I can do this.

The parameter `$digits` only accept integer, so I can't pass argument with text values (`a`, `b`, `c`, `d`, `e`, `f`).

Beside, Rust-core will use the value of this argument to generate the hexadecimal string (e.g. `5f450f4b` in case of `8` digits) at different place, not this place. So no matter how many `digits` this argument is, I can't test that behavior.

Lewiscowles1986: This is my mistake then; as you supply a number of digits and rust generates them. Approved.

Lewiscowles1986 (approved):

PR #386, opened 2023-12-11
refactor: Add RandomBoolean generator class

tienvx:
The generator is defined at https://github.com/pact-foundation/pact-specification/tree/version-4#supported-generators

### DIFF
--- a/src/PhpPact/Consumer/Matcher/Generators/RandomBoolean.php
+++ b/src/PhpPact/Consumer/Matcher/Generators/RandomBoolean.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace PhpPact\Consumer\Matcher\Generators;
+
+use PhpPact\Consumer\Matcher\Model\GeneratorInterface;
+
+/**
+ * Generates a random boolean value
+ */
+class RandomBoolean implements GeneratorInterface
+{
+    /**
+     * @return array<string, string>
+     */
+    public function jsonSerialize(): array
+    {
+        return [
+            'pact:generator:type' => 'RandomBoolean',
+        ];
+    }
+}
--- a/tests/PhpPact/Consumer/Matcher/Generators/RandomBooleanTest.php
+++ b/tests/PhpPact/Consumer/Matcher/Generators/RandomBooleanTest.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace PhpPactTest\Consumer\Matcher\Generators;
+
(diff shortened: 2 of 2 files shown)


src/PhpPact/Consumer/Matcher/Generators/RandomBoolean.php
```diff
@@ -0,0 +1,21 @@
...
+use PhpPact\Consumer\Matcher\Model\GeneratorInterface;
+
+/**
+ * Generates a random boolean value
+ */
+class RandomBoolean implements GeneratorInterface
+{
+    /**
+     * @return array<string, string>
+     */
```
Lewiscowles1986, line 14: thought after looking at several of these, is there not a common format that these fit into?

key, value

tienvx: I'm not sure I understand this question.

Are you suggesting that we can replace `@return array<string, xxx>` with `@return array<string, mixed>` so it's consistent between classes?

Lewiscowles1986: I am suggesting making a data-transfer (readonly public) class to store named arguments, rather than returning an array.

Lewiscowles1986 (approved):

src/PhpPact/Consumer/Matcher/Generators/RandomBoolean.php
tienvx: I can't define a DTO class like this:

```php
class CustomDto
{
    public function __construct(public string $pact:generator:type) {}
}
```

There is a syntax error because of the colon `:`

tienvx: I can prevent returning array by doing this:

```php
public function jsonSerialize(): object
{
        return (object) [
              'pact:generator:type' => 'RandomBoolean',
        ];
}
```

It's a bit hacky, but PHPStan will not complain, and test case is still passed. Then I can remove the `@return` annotation completely.

Is it fine to do this?

tienvx: I created a separate PR for it https://github.com/pact-foundation/pact-php/pull/389

tienvx: Another attempt https://github.com/pact-foundation/pact-php/pull/392

PR #389, opened 2023-12-13
refactor: Remove @return tag from jsonSerialize()

tienvx:
Replace return type from `array` to `object` so no need to add `@return` annotation to explain the return value to make PHPStan happy.

### DIFF
--- a/src/PhpPact/Consumer/Matcher/Generators/AbstractDateTime.php
+++ b/src/PhpPact/Consumer/Matcher/Generators/AbstractDateTime.php
@@ -10,10 +10,7 @@ abstract class AbstractDateTime implements GeneratorInterface
     {
     }
 
-    /**
-     * @return array<string, string>
-     */
-    public function jsonSerialize(): array
+    public function jsonSerialize(): object
     {
         $data = ['pact:generator:type' => $this->getType()];
 
@@ -25,7 +22,7 @@ abstract class AbstractDateTime implements GeneratorInterface
             $data['expression'] = $this->expression;
         }
 
-        return $data;
+        return (object) $data;
     }
 
     abstract protected function getType(): string;
--- a/src/PhpPact/Consumer/Matcher/Generators/MockServerURL.php
+++ b/src/PhpPact/Consumer/Matcher/Generators/MockServerURL.php
@@ -16,12 +16,9 @@ class MockServerURL implements GeneratorInterface
     {
     }
 
-    /**
-     * @return array<string, string>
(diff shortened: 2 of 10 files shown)


tienvx:
One thing to consider: In related PR https://github.com/pact-foundation/pact-php/pull/391 I have to convert data from `object` to `array`, merge with another `array`, then convert the result back to `object`, which is not ideal

Lewiscowles1986:
I'm not sure about this. It feels like a misunderstanding of what I was saying before. While it does have the benefit of named accessed properties, I think just from the comments about PHPStan, it would prefer a user-defined class to StdObject.

tienvx:
Can you give an example how that class look like?

Lewiscowles1986:
The laziest and simplest example of the above could be seen at https://3v4l.org/1utmD

```php
https://3v4l.org/1utmD
```

You could have the metadata's be union types with a method to get metadata

tienvx:
Not sure I understand your suggestion correctly. Here is what I did:

```php
class GeneratorAbstractDTO
{
    public function __construct(
        public string $internalType,
        public string $presentationType,
    ) {
    }
}
```

```php
class RandomBoolean implements GeneratorInterface
{
    public function jsonSerialize(): GeneratorAbstractDTO
    {
        return new GeneratorAbstractDTO(
            'pact:generator:type',
            'RandomBoolean',
        );
    }
}
```

The test failed with 1 failure:

```
1) PhpPactTest\Consumer\Matcher\Generators\RandomBooleanTest::testSerialize
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'{"pact:generator:type":"RandomBoolean"}'
+'{"internalType":"pact:generator:type","presentationType":"RandomBoolean"}'


```

This json is not what Rust-core [expected](https://github.com/pact-foundation/pact-reference/blob/master/rust/pact_ffi/IntegrationJson.md)

Lewiscowles1986:
So when you interface to Rust, it doesn't matter if you use an array; that is an internal detail. But if the design of this is to allow users to interact with this (by users I mean developers, looking to setup Pact); that is where my suggestion fits.

There would need to be something to get the order correct; but that would sit on the internals side.

It looks to me like `jsonSerialize` in this case would still be returning an array though, which might not be ideal. This will be a pig to debug if anyone has to.

tienvx:
Replaced by https://github.com/pact-foundation/pact-php/pull/392. Still using array, not using properties as suggested.

PR #403, opened 2023-12-15
refactor: Add Includes matcher class

tienvx:
The matcher is defined at https://github.com/pact-foundation/pact-specification/tree/version-4#supported-matching-rules


Lewiscowles1986 (approved):
A Note for the PACT team that if to explain includes, you use contains. Contains is a better word!

tienvx:
I also agree that `Contains` is better word.

We can discuss about this in Slack later.

For now I would like to keep this `Includes` word to make it consistent with the table in the specification.

PR #391, opened 2023-12-13
refactor: Add Regex matcher class

tienvx:
The matcher is defined at https://github.com/pact-foundation/pact-specification/tree/version-4#supported-matching-rules

Depend on https://github.com/pact-foundation/pact-php/pull/392

PR #415, opened 2023-12-15
refactor: Add Values matcher class

tienvx:
The matcher is defined at https://github.com/pact-foundation/pact-specification/tree/version-4#supported-matching-rules

### DIFF
--- a/src/PhpPact/Consumer/Matcher/Matchers/Values.php
+++ b/src/PhpPact/Consumer/Matcher/Matchers/Values.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace PhpPact\Consumer\Matcher\Matchers;
+
+use PhpPact\Consumer\Matcher\Model\MatcherInterface;
+
+/**
+ * Match the values in a map, ignoring the keys
+ */
+class Values implements MatcherInterface
+{
+    /**
+     * @param array<mixed> $values
+     */
+    public function __construct(private array $values)
+    {
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    public function jsonSerialize(): array
+    {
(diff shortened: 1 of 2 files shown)


Lewiscowles1986:
There are over 20 of these. Would you mind adding a bit more detail, or perhaps letting me know expected order or priority?

Some of these feel like things I may have approved before.

tienvx:
* There are 12 classes for generators.
  * You **reviewed all** of them.
  * I removed `jsonSerialize` methods for all of them. Look like they are not needed.
* There are 23 classses for matchers.
  * You **reviewed 2** of them (`Equality` and `Regex`).
  * The method `jsonSerialize` is required for these classes to work.
  * You can review the pull request in **any order.** There are 2 notes:
    * It also applied to pull requests with **failed test case**
    * Only **one exception**:
      * First `Date` #394, then
        * `Time` #395
        * Or `DateTime` #396
  * If the class extends `GeneratorAwareMatcher`, you can see that class here https://github.com/pact-foundation/pact-php/pull/391/files#diff-bb4df63ac65cc79b12dfada6442c909439b9f92b26bdd54a197cdd5237cdc84c

Lewiscowles1986 (approved):

Lewiscowles1986:
Do you think once these merge you might be able to write up a sample usage difference of before and after?

tienvx:
I already prepare examples (also act as integration tests) for:
* [matchers](https://github.com/pact-foundation/pact-php/tree/ffi/example/matchers)
* [generators](https://github.com/pact-foundation/pact-php/tree/ffi/example/generators)

Once all PRs are merged, I expect there will be **no** difference for users to use them.

PR #417, opened 2023-12-18
chore: Print received message's metadata

tienvx:


Lewiscowles1986 (approved):
So one thing that would be nice is if there were a flag to guard this (future, non-blocking suggestion); but personally I prefer verbosity over the silence I saw when I first started using Pact with PHP

tienvx:
I created a ticket for this kind of improvements https://github.com/pact-foundation/pact-php/issues/419

PR #420, opened 2023-12-18
feat: Throw exception when both generator and example value are set

tienvx:
Generator can't be used along with example value:

- Generators (except `ProviderState`) will be silently ignored (without any notice).
- `ProviderState` generator will be used and example value will be ignored (without any notice)

User will be surprised by this.

This PR will throw exception instead.

These code won't be allowed:
- `$this->matcher->fromProviderState($this->matcher->integer(123), '${id}')` (user won't know, until the pact is verified on provider side. And user probably won't notice that `123` is used instead the number from provider)
- `$matchingRule = $this->matcher->integerV3(123); $matchingRule->setGenerator(new RandomInt());` (user probably won't notice that `123` is used instead of a random integer.)

### DIFF
--- a/src/PhpPact/Consumer/Matcher/Exception/GeneratorNotRequiredException.php
+++ b/src/PhpPact/Consumer/Matcher/Exception/GeneratorNotRequiredException.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace PhpPact\Consumer\Matcher\Exception;
+
+class GeneratorNotRequiredException extends MatcherException
+{
+}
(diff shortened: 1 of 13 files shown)


Lewiscowles1986 (approved):
Does a bit more than PR title suggests, but all looks okay. I prefer the test-case-local to the test-suite-file class-local

tienvx:
> I prefer the test-case-local to the test-suite-file class-local

Are you talking about this base class `GeneratorAwareMatcherTestCase` ?

Can you explain a bit more? And what is your approach? I'd love to hear it.

Lewiscowles1986:
> I prefer the test-case-local to the test-suite-file class-local 

This was a compliment on your refactor.

PR #423, opened 2023-12-19
chore: Remove old composer script

tienvx:
This script is from version <=9.

In version 10 I refactored the example/ directory.

The public path in the script is not correct any more. And I don't think this script is useful. So I removed it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -73,7 +73,6 @@
         }
     },
     "scripts": {
-        "start-provider": "php -S localhost:58000 -t example/src/Provider/public/",
         "static-code-analysis": "phpstan",
         "lint": "php-cs-fixer fix --dry-run",
         "fix": "php-cs-fixer fix",


Lewiscowles1986:
Is it a case that there has been a breakage in this script in the refactor?

I'm all for removing it if the pipeline passes (which it does); but I'd just like to understand a little more.

tienvx:
Here is the old [index.php](https://github.com/pact-foundation/pact-php/blob/master/example/src/Provider/public/index.php) file.
Here is the new [index.php](https://github.com/pact-foundation/pact-php/blob/ffi/example/json/provider/public/index.php) file.

I don't change it much. I think the script is still working fine (just change the public's path).

But I think it's not useful. I can't think of any reason the script must be there.

tienvx:
I tried remove that script and run tests on `master` branch and all tests are passed.

Lewiscowles1986 (approved):

PR #424, opened 2023-12-20
test(compatibility-suite): Add behat

tienvx:


Lewiscowles1986 (changes requested):
please close this as standalone it's too small, even for me.

PR #433, opened 2023-12-24
test(compatibility-suite): Add PactPath DTO

tienvx:


compatibility-suite/tests/Model/PactPath.php
```diff
@@ -0,0 +1,32 @@
...
+
+    public function getConsumer(): string
+    {
+        return $this->consumer;
+    }
+
+    public function getProvider(): string
+    {
+        return self::PROVIDER;
+    }
```
Lewiscowles1986, line 24: Can you please choose a consistent way of defining?

I was thinking private readonly strings for both provider and consumer.

tienvx: Removed the method `getProvider`. Now only use the constant.

We didn't remove php 8.0 yet so the keyword `readonly` is not supported. The property `$consumer` is read only but without the keyword `readonly`.

Lewiscowles1986 (approved):

PR #435, opened 2023-12-29
chore: Check string generator is working

tienvx:
I think `RandomString` generator work a bit different from other generators. It work even if the `value` is still there. But it's totally fine. We want this behavior, because:
1. There is no matcher for `string`
2. We need to re-use `type` matcher with a string value
3. That's why `value` must be always a string, even though we are using the `RandomString` generator or not.


Lewiscowles1986 (approved):
Also consider some other asserts, such as not null, not undefined, length

PR #442, opened 2024-01-07
refactor!: Move protobuf code to core

tienvx:
- Move code from [protobuf](https://github.com/tienvx/pact-php-protobuf) to this repo
- Only move the code, there are 2 examples for protobuf and will be in separated PRs
- No tests for these code yet

PR #444, opened 2024-01-09
chore: Add protobuf sync message example

tienvx:
Depend on https://github.com/pact-foundation/pact-php/pull/442

- Move example from [protobuf](https://github.com/tienvx/pact-php-protobuf) to this repo
- Only example for `sync message`. Example for `async message` is in [another PR](https://github.com/pact-foundation/pact-php/pull/443)
- Tests are slow significantly
  - On Linux arm, it used to be ~ 30 minutes, now it's ~ 8 minutes
  - On Macos arm instance, it used to be ~ 10 minutes, now it's ~ 5 minutes
  - Look like number of cpu/ram on macos instance are always 4/12G, so I would like to hard code the number of jobs to 4. I think `MAKEFLAGS=" -j$(nproc)"` doesn't work on macos https://cirrus-ci.com/task/6008438270984192?logs=pre_req#L259. `MAKEFLAGS=" -j $(nproc)"` (a space between `-j` and `$(nproc)`) does not work either https://cirrus-ci.com/task/5204886841720832
  - I can't change number of cpu/ram for macos instance https://cirrus-ci.org/guide/macOS/
  - I can change number of cpu/ram for linux arm container, so I change it to the same value as macos instance. 4G ram is not enough for 4 jobs https://cirrus-ci.com/task/4885518131920896
  - I would like to hard code the number of jobs to 4 as well. Here is an comment that I consider use hard code number https://github.com/grpc/grpc/issues/23626#issuecomment-1416121110
  - I will improve compiling speed by using this [feature](https://cirrus-ci.org/guide/docker-builder-vm/#dockerfile-as-a-ci-environment) from Cirrus CI in the next PR
- On Windows, `shivammathur/setup-php` can't install `grpc` extension. I can't build it manually, no matter I tried. So I look at this [forum](https://www.apachelounge.com/viewtopic.php?p=42339), grab a [build](https://phpdev.toolsforresearch.com/php-8.2.7-nts-Win32-vs16-x64-grpc-protobuf.zip) and extract grpc extension. This is the only way while waiting for someone to fix this [issue](https://externals.io/message/117494#119568)


example/protobuf-sync-message/provider/composer.json
```diff
@@ -0,0 +1,30 @@
+{
+    "name": "pact-foundation/example-protobuf-sync-message-provider",
+    "require": {
+        "grpc/grpc": "^1.57",
+        "spiral/roadrunner-grpc": "^2.0",
+        "tienvx/composer-downloads-plugin": "^1.2"
+    },
+    "require-dev": {
+        "ext-grpc": "*"
```
tienvx, line 9: This line should be in `require`, not `require-dev`. But if it is in `require`, `php-cs` and compatibility suite jobs will need to install `grpc` extension, which make them slow significantly

Lewiscowles1986: What about using `"suggest"` in  composer.json?

```
"suggest": {
    "ext-grpc": "If you want to use gRPC then you must install and use this"
}
```

Lewiscowles1986:
I was actually getting involved in windows builds for PHP around July, but left interest, because of complete misery interacting with some core PHP maintainers (seemed to be against iterative work)

example/protobuf-sync-message/provider/composer.json
tienvx: Updated.

The suggestion is hidden in this message:

```
18 package suggestions were added by new dependencies, use `composer suggest` to see details.
```

I have to run this command  `composer suggest` to see it clearly. But it's better than nothing.

Lewiscowles1986: Sorry, I think I should have been clearer.

The package should exist within `require-dev` as well as `suggest`:

- Suggest presence is to let devs know that if they want gRPC they should install the extension
- Require-dev presence is to clearly communicate that to run the test suite, or work on this project, because tests contain gRPC, they will need to find a way to install that package.

tienvx: Reverted. Now **this** `composer.json` contains both `require-dev` and `suggest`. I'm not sure which is better:  move `suggest` to the **root** `composer.json`, or keep it here.

Lewiscowles1986 (approved):
👍

PR #445, opened 2024-01-13
ci(test): Compile grpc and cache Linux docker image on Cirrus CI

tienvx:
Depend on https://github.com/pact-foundation/pact-php/pull/444

Only apply to linux. The task is now run in `30 seconds` not `30 minutes` (with the same hardware: 2 cores, 4G ram).

Can't apply to macos, because of this error `Intel macOS instances are no longer supported!` https://cirrus-ci.com/task/6049929467002880. I tried [another approach](https://github.com/pact-foundation/pact-php/pull/446) but it doesn't work

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -27,17 +27,9 @@ linux_arm64_task:
       - VERSION: 8.1
       - VERSION: 8.0
   arm_container:
-    image: php:$VERSION
-    cpu: 4
-    memory: 12G
-  pre_req_script: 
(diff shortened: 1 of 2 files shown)


YOU54F:
yeah unfortunatly macos with cirrus uses virtualization (apple's hypervisor framework) which has a limitation of no nested virtualisation so docker isn't available on the macos configurations on cirrus.

It also isn't available for macos, with GitHub actions. I believe the only free option that supports Docker + Mac is CircleCI

YOU54F (approved):

PR #448, opened 2024-01-16
refactor: Add PluginFormatter

tienvx:
* Add `MatchingField` matcher
* Add `PluginFormatter`
* Allow customize formatter for `Matcher` helper class
* Apply it to plugin examples e.g. Replace `"matching(type,'Name')"` by `$matcher->like('Name')`
* Syntax and supported matchers of `matching` expression: https://github.com/pact-foundation/pact-plugins/blob/main/docs/matching-rule-definition-expressions.md
* All matchers: https://github.com/pact-foundation/pact-specification/tree/version-4#supported-matching-rules

### DIFF
--- a/example/csv/consumer/tests/Service/HttpClientServiceTest.php
+++ b/example/csv/consumer/tests/Service/HttpClientServiceTest.php
@@ -4,6 +4,8 @@ namespace CsvConsumer\Tests\Service;
 
 use CsvConsumer\Service\HttpClientService;
 use PhpPact\Consumer\InteractionBuilder;
+use PhpPact\Consumer\Matcher\Formatters\PluginFormatter;
+use PhpPact\Consumer\Matcher\Matcher;
 use PhpPact\Consumer\Model\Body\Text;
 use PhpPact\Consumer\Model\ConsumerRequest;
 use PhpPact\Consumer\Model\ProviderResponse;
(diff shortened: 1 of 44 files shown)


Lewiscowles1986:
So one concern with these works, is that anyone using your work, might see backwards compatibility breaks. I Don't think this should be blocked entirely, but I do know that at some point, you'll want a stable interface for pact users and plugin authors alike.

Do you have some form of view around the time-frame, and might it be advantageous to use some form of phpDoc comment to annotate?

tienvx:
A remind for @Lewiscowles1986

Lewiscowles1986:
Ah, I'd missed the reply here. 

RE:

> > Do you have some form of view around the time-frame, and might it be advantageous to use some form of phpDoc comment to annotate?

> Can you rephrase this question for me.

Do you know how long it will take for this to come out of alpha?

Really as you've quantified one user impact, it should not matter anyway.

Lewiscowles1986 (approved):

tienvx:
> Do you know how long it will take for this to come out of alpha?

I have some items in my mind for the next 2 or 3 more alpha versions:
* Refactor code related to InteractionBuilder/MessageBuilder if needed
* Add unit tests as much as possible
* Update document/readme

Probably it will take 1 or 2 months

PR #450, opened 2024-01-23
ci: Reduce test running time by setting PACT_DO_NOT_TRACK

tienvx:
* On local, where everything are set up, it reduce tests running time, from `6.549` seconds to `4.242` (need to set manually `PACT_DO_NOT_TRACK=true phpunit`)
* On github, it's harder to see the difference

### DIFF
--- a/.github/workflows/compatibility-suite.yml
+++ b/.github/workflows/compatibility-suite.yml
@@ -3,7 +3,7 @@ name: Pact-PHP Compatibility Suite
 on: [push, pull_request]
 
 env:
-  pact_do_not_track: true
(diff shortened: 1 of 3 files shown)


.github/workflows/compatibility-suite.yml
```diff
@@ -3,7 +3,7 @@ name: Pact-PHP Compatibility Suite
 on: [push, pull_request]
 
 env:
-  pact_do_not_track: true
+  PACT_DO_NOT_TRACK: true
```
tienvx, line 6: It's the same. Just for consistency

YOU54F (approved):

PR #451, opened 2024-01-23
chore: Reuse process handling code

tienvx:
* Rename `ProviderProcess` to `AbstractProcess` to reuse code
* Add `PhpProcess`
* Add `RoadRunnerProcess`
* Use dynamic port for `PhpProcess`


Lewiscowles1986 (approved):
sockets are one of the first things I look for as a warning,  but this simplifies the calling to be in one place. Nice one @tienvx

Issue #419, opened 2023-12-18
Pact PHP v10 > Add ability to silence output when running unit tests


tienvx:
Currently when running `composer test`, a lot of debug output is printed into the console.

It'd be nice if there is a flag to disable all of debug output.

It's expected to print only `.` (dots) when turned that flag.

PR #452, opened 2024-01-24
chore: Only show logs if configured

tienvx:
Depend on #451

For #419

* Don't show logs by default (when running `phpunit`)
* FFI: show logs based on value of `PACT_LOGLEVEL=trace|debug|info|warn|error`
* Provider processes: show logs if `PACT_LOGLEVEL` is set to any value (because I don't know which level is the best to start showing these logs)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -23,6 +23,7 @@ linux_arm64_task:
   env:
     COMPOSER_ALLOW_SUPERUSER: 1
     PACT_DO_NOT_TRACK: true
+    PACT_LOGLEVEL: debug
     matrix:
       - VERSION: 8.2
       - VERSION: 8.1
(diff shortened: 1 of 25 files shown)


Lewiscowles1986:
I actually prefer log level to show me everything in CI when consuming the pact-php library and locally; but I can see why some devs might not like the noise. For me it highlights when I'm making silly mistakes.

tienvx:
Currently `PACT_LOGLEVEL=none` is not working correctly (I remember core team even consider to drop support for it). So currently, disable logs by default and allow users to manually enable it is the supported way by core team, not the other way around.

tienvx:
Let me try to fix `PACT_LOGLEVEL=none` then revert the code

tienvx:
Replaced by #454

PR #459, opened 2024-02-05
chore: Throw CDataNotCreatedException

tienvx:


Lewiscowles1986:
Can you describe this a bit more for me. I Know what it is doing, but not why. Is there an exception otherwise in the Rust code which is using FFI ?

tienvx:
> Can you describe this a bit more for me. I Know what it is doing, but not why. 

* The reason for this PR: I just want to make it consistent between `ArrayData` and `BinaryData`
* The reason for this code: I want to replace this type error by the exception:
```
PHP Fatal error:  Uncaught TypeError: memcpy(): Argument #1 ($to) must be of type CData, null given
```

I think it's easier to debug (I could be wrong, maybe this type error is clearer than my custom exception) on failure (PHP can't create C data structure as requested - I don't know how to test it though)

> Is there an exception otherwise in the Rust code which is using FFI ?

There is no other exception. It's not related to Rust. It's just pure PHP code.

Lewiscowles1986 (approved):

Lewiscowles1986:
Thank you for your explanation. I Do find that those work very well in the summary box to avoid such questions, but thank you all the same. Approved.

PR #461, opened 2024-02-06
chore: Deprecate Values matcher

tienvx:
`Values` matcher has weird behaviors. This PR follow the same approach from `pact-js`: `Values` matcher is deprecated in favour of `EachKey` and `EachValue` matchers

See https://github.com/pact-foundation/pact-reference/pull/365


Lewiscowles1986 (approved):
Stops people learning about deprecated code from examples. Nice one

PR #468, opened 2024-02-13
deps: Upgrade PHP to 8.3 and drop 8.0

tienvx:
* Upgrade PHP to 8.3
* Drop PHP 8.0
* I was able to compile grpc extension for PHP 8.2 & 8.3 on Windows, so I switch to that file
* Tests on PHP 8.2 and 8.3 on Windows are easier to fail than before with this error `Script phpunit --debug handling the test event returned with error code -1073741819`, I'm not sure why

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         operating-system: [ ubuntu-latest, windows-latest, macos-12, macos-14 ]
-        php: [ '8.0', '8.1', '8.2' ]
+        php: [ '8.1', '8.2', '8.3' ]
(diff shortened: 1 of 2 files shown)


Lewiscowles1986 (approved):
Thought lock was missing, but apparently this has not lock file...

Can the examples be updated in separate or once release is cut please?

tienvx:
> Thought lock was missing, but apparently this has not lock file...

I think it make sense, according to https://getcomposer.org/doc/02-libraries.md#lock-file

> Can the examples be updated in separate or once release is cut please?

I don't understand this question. Are you talking about these [examples](https://github.com/pact-foundation/pact-php/tree/ffi/example) ? Those examples don't have `composer.lock` so no need to update anything.

PR #470, opened 2024-02-14
Use pre-built grpc extension

tienvx:
* Use pre-built grpc extension on all systems to speed up test jobs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,18 +40,18 @@ jobs:
         run: composer run static-code-analysis
 
   test:
-    runs-on: ${{ matrix.operating-system }}
+    runs-on: ${{ matrix.os }}
     needs:
       - php-cs
     strategy:
       fail-fast: false
       matrix:
-        operating-system: [ ubuntu-latest, windows-latest, macos-12, macos-14 ]
+        os: [ ubuntu-22.04, windows-2022, macos-12, macos-14 ]
         php: [ '8.1', '8.2', '8.3' ]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -60,23 +60,35 @@ jobs:
...
+            ext=dll
+          else
+            ext_dir=$(php-config --extension-dir)
+            ini_path=$(php-config --ini-dir)/grpc.ini
+            sudo=sudo
+            lib_name=grpc.so
+            ext=so
+          fi
+          $sudo curl -L -o $ext_dir/$lib_name https://github.com/tienvx/build-php-grpc-extension/releases/download/builds/grpc_${{ matrix.os }}_php-${{ matrix.php }}.$ext
+          echo "extension=$lib_name" | $sudo tee -a $ini_path
```
Lewiscowles1986, line 90: I have concerns about this. Also why all the renaming of operating-system to os?

I think a pre-requisite is that the grpc extension come into the PACT Foundation and is audited by someone that understands what that binary blob does.

tienvx: > Also why all the renaming of operating-system to os?

Because it's shorter when it's repeated multiple times in the PR.

> I think a pre-requisite is that the grpc extension come into the PACT Foundation and is audited by someone that understands what that binary blob does.

The code is open source at https://github.com/tienvx/build-php-grpc-extension.
I think it's not worth continue doing this. I will close this PR.
I will also remove this github action step when grpc extension binary for PHP 8.2 & 8.3 on windows is available again.

PR #471, opened 2024-02-14
Upgrade phpunit

tienvx:
* Since php is upgraded to ^8.1, phpunit can be upgrade to ^10
* `assertRegExp` is no longer available in phpunit 10

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,7 +62,7 @@ jobs:
         with:
           extensions: ${{ matrix.operating-system == 'windows-latest' && contains('["8.2","8.3"]', matrix.php) && 'sockets, curl, zip, ffi' || 'sockets, curl, zip, ffi, grpc' }}
           php-version: ${{ matrix.php }}
-          coverage: none
+          coverage: pcov
           ini-values: ${{ matrix.operating-system == 'windows-latest' && 'opcache.enable=0 opcache.enable_cli=0' || '' }}
 
       - name: Install Protoc
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Generate Library
         run: composer gen-lib
 
-      - name: Run Tests
-        run: vendor/bin/phpunit --debug
+      - name: Composer test
+        run: composer test
         env:
           PACT_DO_NOT_TRACK: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Generate Library
         run: composer gen-lib
 
-      - name: Composer test
-        run: composer test
+      - name: Run Tests
+        run: vendor/bin/phpunit --debug
         env:
           PACT_DO_NOT_TRACK: true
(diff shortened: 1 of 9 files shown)


Lewiscowles1986 (changes requested):
can you remove the grpc binary edits from this PR please. Also tests seem to be failing. Should this be a draft PR?

Lewiscowles1986:
> Unknown option "--debug"

I wonder if a shortcut might be `--verbose` flag?

https://github.com/sebastianbergmann/phpunit/issues/1703#issuecomment-97677445

Lewiscowles1986:
See also https://github.com/sebastianbergmann/phpunit/issues/4704#issue-915089232

.github/workflows/build.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         with:
           extensions: ${{ matrix.operating-system == 'windows-latest' && contains('["8.2","8.3"]', matrix.php) && 'sockets, curl, zip, ffi' || 'sockets, curl, zip, ffi, grpc' }}
           php-version: ${{ matrix.php }}
-          coverage: none
+          coverage: pcov
```
Lewiscowles1986, line 65: I Have a feeling, that this is if x-debug is unavailable that phpunit falls back to requiring pcov; I can't remember the specifics about why.

tienvx:
> I wonder if a shortcut might be --verbose flag?

Fixed by upgrade phpunit to at least `10.5.6`

.github/workflows/build.yml
tienvx: xdebug/pcov is required by phpunit, otherwise we got this error:

> Script phpunit --debug handling the test event returned with error code 1

pcov is faster so I choose it

composer.json
```diff
@@ -96,10 +96,7 @@
         "static-code-analysis": "phpstan",
         "lint": "php-cs-fixer fix --dry-run",
         "fix": "php-cs-fixer fix",
-        "test": [
-            "php -r \"array_map('unlink', glob('./example/*/pacts/*.json'));\"",
-            "phpunit --debug"
-        ],
+        "clean-up-pacts": "php -r \"array_map('unlink', glob('./example/*/pacts/*.json'));\"",
```
Lewiscowles1986, line 99: This will need to be documented as a breaking dependency. @YOU54F this change looks good; but how does PACT, or even PACT-PHP usually deal with this?

Also why would the process not be able to delete these files on windows? It feels like perhaps there is a locking bug?

Lewiscowles1986:
This is the coverage making it fail on windows... https://github.com/pact-foundation/pact-php/pull/472 uses this. I basically checked a few different things. I don't think it's impossible to get pcov working under windows, but also it's not the core purpose of this PR; so I'd likely add it to some backlog of "nice-to-have" features.

Nice work

composer.json
tienvx: Reverted

tienvx:
pcov did work on my machine. Not sure why it causing problems on github actions.

This annoying error still happen:

> Script phpunit --no-coverage handling the test event returned with error code -1073741819

I will focus on that next.

Thanks for helping me solving the errors.

Lewiscowles1986 (approved):

PR #472, opened 2024-02-15
Chore/php unit 10 upgrade

Lewiscowles1986:
just testing, so I'm not pushing quite so much onto tienvx


Lewiscowles1986:
https://stackoverflow.com/questions/63063990/laravel-errorexception-hash-file-read-of-8192-bytes-failed-with-errno-13-pe and https://stackoverflow.com/questions/8810094/phpunit-memory-limit-parameter-does-not-apply suggested playing with memory limit. I'd experimented with switching from pcov to xdebug and upping memory limit to 4GB... Now I'm going to see if I can just run without coverage.

PR #474, opened 2024-02-15
deps: Update roadrunner binary

tienvx:
Update roadrunner binary in the hope of fixing the error `Script phpunit --no-coverage handling the test event returned with error code -1073741819` but it couldn't


Lewiscowles1986 (approved):
a newer version shouldn't harm anyone

Issue #206, opened 2021-07-22
Add support for consumer version selectors for deployed and released versions


bethesque:
Please add support for the following keys to be used in the consumer version selectors when fetching pacts for verification:

```
{ "deployedOrReleased": true }

{ "deployed": true }

{ "released": true }

{ "environment": "<env name>" }

{ "consumer": "<name>" } // this may already be supported, please add it if not
```

These keys can be used in various combinations ( eg. `{ "environment": "prod", released: true, consumer: "Foo" }` ). Please allow any combination of keys, and do not validate them locally, as the validation rules can change over time as support is added for new selectors - just ensure that any errors returned from the API are displayed clearly to the user.

See https://github.com/pact-foundation/pact_broker/blob/master/lib/pact_broker/doc/views/provider-pacts-for-verification.markdown for the usage docs of each selector.

As more selectors will be added over time, it's my preference that the consumer version selectors are not strongly typed, and are just passed through straight from the configuration object to the API, so that we don't have to make code changes to the Pact clients each time we add new selectors. (eg. there will be new selectors coming for branch support very soon). I understand that this is not compatible with the idioms of every Pact client language however, so do whatever you need to do to make the new selectors work elegantly within your language.

Please update https://docs.pact.io/pact_broker/advanced_topics/consumer_version_selectors/ for your language once the selectors have been released.

bethesque:
What is the status on this please? I'm updating the docs with a section on which languages/versions support the selector, so people know if they can use it or not.

YOU54F:
arbitrary json can be provided in 9.x.x for the selectors

feature was introduced in https://github.com/pact-foundation/pact-php/pull/213

@tienvx has implemented in the v10.x alpha release, and main branch to also validate the json provided are valid selectors. 

docs

https://github.com/pact-foundation/pact-php/blob/master/docs/provider.md#verify-from-pact-broker

and in code, here is a test showing named selectors, and json selectors being passed.

https://github.com/pact-foundation/pact-php/blob/912b5f4831439f392de6816bfa969e728c0ad52d/tests/PhpPact/Standalone/ProviderVerifier/Model/ConsumerVersionSelectorsTest.php#L15-L17

Might not meet the brief on this

> Please allow any combination of keys, and do not validate them locally, as the validation rules can change over time as support is added for new selectors - just ensure that any errors returned from the API are displayed clearly to the user.

[...]

PR #477, opened 2024-02-16
feat: Add Selector class

tienvx:
Instead of asking user to work with json directly, user only need to initiate the object of `Selector` class using PHP 8 named arguments feature. This class help validate and convert selector to json.

Related issue: https://github.com/pact-foundation/pact-php/issues/206
Docs: https://docs.pact.io/pact_broker/advanced_topics/consumer_version_selectors


Lewiscowles1986 (approved):
I had never seen this form of spread before, but I'm here for it. Very clever.

PR #478, opened 2024-02-16
chore: Remove unused exception ConnectionException

tienvx:
The exception `ConnectionException` is for version <=9 when user work with mock server directly.
At version 10, mock server is automatically started and stopped, so this exception is no longer needed


Lewiscowles1986:
Approved, but logically, this speaks to a degredation in code quality as we can't discern specific classes of errors.

Lewiscowles1986 (approved):

tienvx:
> At version 10, mock server is automatically started and stopped, so this exception is no longer needed

I was wrong about this. I mean:

> At version 10, user doesn't work with mock server directly, so this exception is no longer needed

Issue #371, opened 2023-11-30
FFI > Free unmanaged memory when no longer used


tienvx:
There are 2 places that create unmanaged memory:

- https://github.com/pact-foundation/pact-php/blob/ffi/src/PhpPact/FFI/Model/ArrayData.php#L44
- https://github.com/pact-foundation/pact-php/blob/ffi/src/PhpPact/FFI/Model/BinaryData.php#L34

There is a potential memory leak if it's not handled.

- I already call `FFI::free` for the first one (https://github.com/pact-foundation/pact-php/blob/ffi/src/PhpPact/FFI/Model/ArrayData.php#L58), but I'm not sure it's correctly handled.
- I don't know solution for the second one

tienvx:
* `FFI::free` work fine in `ArrayData`
* I changed from `unmanaged` data to `managed` data in `BinaryData`

YOU54F:
Is this resolved now @tienvx

tienvx:
I think it's resolved

PR #481, opened 2024-02-17
fix: FFI > Create managed binary data

tienvx:
FFI: Change from unmanaged data to managed data

For #371

### DIFF
--- a/src/PhpPact/Consumer/Model/Body/Binary.php
+++ b/src/PhpPact/Consumer/Model/Body/Binary.php
@@ -10,6 +10,8 @@ class Binary
 {
     use ContentTypeTrait;
 
+    private ?BinaryData $data = null;
+
     public function __construct(private string $path, string $contentType)
     {
         $this->setContentType($contentType);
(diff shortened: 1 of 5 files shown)


Lewiscowles1986 (approved):
Would appreciate any links you have on this, but I think it should help to prevent memory leaks if managed is the same type of [memory] management I'd think of. Interesting that it's not the default as past ffi, it's in PHP and should work as PHP would expect (managed)

tienvx:
Here is the link about `$owned` parameter of `FFI::new()`:

https://www.php.net/manual/en/ffi.new.php#refsect1-ffi.new-parameters

> owned: Whether to create owned (i.e. managed) or unmanaged data. Managed data lives together with the returned [FFI\CData](https://www.php.net/manual/en/class.ffi-cdata.php) object, and is released when the last reference to that object is released by regular PHP reference counting or GC. Unmanaged data should be released by calling [FFI::free()](https://www.php.net/manual/en/ffi.free.php), when no longer needed.

PR #483, opened 2024-02-19
refactor: Add InteractionPart enum

tienvx:
This code will be used later

Use case:

```php
$driver->registerBody($interaction, InteractionPart::REQUEST);
$driver->registerBody($interaction, InteractionPart::RESPONSE);

public function registerBody(Interaction $interaction, InteractionPart $part): void
{
    $body = $part->isRequest() ? $interaction->getRequest()->getBody() : $interaction->getResponse()->getBody();
    ...
}
```


Lewiscowles1986 (commented):
Can you let me know a little more about the intended use-case for this?

Lewiscowles1986:
This feels like it's compensating for a lack of a getBody method, which should not care if the object is a request or a response.

tienvx:
Added method `getBody` and `getHeaders` to `Interaction`

tienvx:
Ping @Lewiscowles1986 . I think you miss this one

Lewiscowles1986 (approved):
Hey @tienvx I'll approve to unblock as I can't see any deep problems and I'd like to unblock you; but it likely still needs iteration. The enum, the use of match, are papering over design problems.

I was trying to work out how best to communicate without just writing code. I won't do that as it seemed like it might be obnoxious on my part.

tienvx:
Thank you. If you have additional ideas please continue comment to this discussion

PR #484, opened 2024-02-19
refactor: Add InteractionPartUsingPlugin enum

tienvx:
This code will be used later


tienvx:
> This doesn't feel like an enum here

It can be 3 different constants. But I think enum is better because I can put related code into a same place.

tienvx:
I am thinking about another approach that doesn't need extra enum or class. Closing.

PR #485, opened 2024-02-19
refactor: Add handle trait

tienvx:
This is will be used later


src/PhpPact/Consumer/Model/Interaction/IdTrait.php
```diff
@@ -0,0 +1,20 @@
...
+namespace PhpPact\Consumer\Model\Interaction;
+
+trait IdTrait
+{
+    private int $id;
+
+    public function getId(): int
+    {
+        return $this->id;
+    }
```
Lewiscowles1986, line 12: is there a reason that the trait needs to store as int? I feel like IntIdTrait makes most sense here, as most synthetic primary keys I use are not integers at all.

tienvx: Renamed `id` to `handle`, as how it's called in Rust

PR #487, opened 2024-02-21
refactor: Add FFI\Lib to prevent using FFI\Client directly

tienvx:
```
$this->lib->getInteractionPartId(InteractionPart::REQUEST);
```

will replace

```
$this->ffi->get('InteractionPart_Request');
```

Please suggest better name if `Lib` is not good one.

Depend on https://github.com/pact-foundation/pact-php/pull/483


tienvx:
Closed as I don't want to refactor this code right now.

PR #490, opened 2024-02-24
refactor: Add interaction part drivers

tienvx:
* Add part drivers (code taken from part registry)
* Add tests for part drivers

### DIFF
--- a/src/PhpPact/Consumer/Driver/InteractionPart/AbstractInteractionPartDriver.php
+++ b/src/PhpPact/Consumer/Driver/InteractionPart/AbstractInteractionPartDriver.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace PhpPact\Consumer\Driver\InteractionPart;
+
+use PhpPact\Consumer\Driver\Body\InteractionBodyDriver;
+use PhpPact\Consumer\Driver\Body\InteractionBodyDriverInterface;
+use PhpPact\Consumer\Driver\Enum\InteractionPart;
+use PhpPact\Consumer\Model\Interaction;
+use PhpPact\FFI\ClientInterface;
+
+abstract class AbstractInteractionPartDriver
+{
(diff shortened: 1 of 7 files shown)


src/PhpPact/Consumer/Driver/InteractionPart/AbstractInteractionPartDriver.php
```diff
@@ -0,0 +1,44 @@
...
+        ?InteractionBodyDriverInterface $bodyDriver = null
+    ) {
+        $this->bodyDriver = $bodyDriver ?? new InteractionBodyDriver($client);
+    }
+
+    protected function withBody(Interaction $interaction, InteractionPart $part): self
+    {
+        $this->bodyDriver->registerBody($interaction, $part);
+
+        return $this;
```
Lewiscowles1986, line 26: This feels like it enforces, or expects a sequence of calls to initialize an object to a state ready for use.
`withBody` and `withHeaders` could be available, but should be initializable without these calls, I think. I won't block on it, but it could manifest in folks becoming confused.

Lewiscowles1986 (approved):
I won't block on it, but the mutable `return $this`; in what seem like objects borrowing from the builder pattern, can be a source of bugs.

I've also noted the difficulty with instantiating an object that needs it's body and headers to be coordinated like this.

src/PhpPact/Consumer/Driver/InteractionPart/AbstractInteractionPartDriver.php
tienvx: I will check and remove `$this` if these methods can be call in any order in a later PR

tienvx: PR https://github.com/pact-foundation/pact-php/pull/492

PR #497, opened 2024-02-29
ci: Split jobs

tienvx:
Apply multiple changes to fix **random fails**:

* Update PHPUnit's config file `phpunit.xml`:
  * Reduce number of suites
  * Change suite's names to make it easier to reference in steps below.
* Default PHP version is used to speed up the jobs. According to this [table](https://github.com/shivammathur/setup-php?tab=readme-ov-file#cloud-osplatform-support):
  * Ubuntu: `8.1`
  * MacOS 12: `8.2`
  * Windows: `8.2`
  * MacOS 14 (ARM): currently not supported. Assume `8.2`
* Add `unit` job:
  * No need to run on multiple os, only `ubuntu` is enough
  * Need to run on multiple php versions `8.1 -> 8.3`
  * Need to run on multiple dependency configurations `lowest/locked`
  * Can now enable code coverage checking
* Rename `test` job to `examples`
  * No need to run on multiple php versions, single version is enough
  * No need to run on multiple dependency configuration, `locked` is fine
  * Ubuntu/MacOS:
    * Run all examples at once
    * Need `grpc` extension
    * Need `protoc` tool
  * **Windows**:
    * Run examples on **parallel** to fix **random fails**
    * `protobuf-sync-message` suite
      * Need `grpc` extension
      * Need `protoc` tool
      * Need PHP `8.1`
    * `protobuf-async-message` suite
      * No need `grpc` extension
      * Need `protoc` tool
      * Use default PHP version `8.2` to improve speed
    * Other suites
      * Use default PHP version `8.2` to improve speed
      * No need `grpc` extension
      * No need `protoc` tool
* Update `php-cs` job
  * Use default PHP version `8.1`
* MacOS 14 (ARM) still got this random error `dyld[7148]: Library not loaded: /opt/homebrew/opt/xz/lib/liblzma.5.dylib`, probably because it's not supported by `shivammathur/setup-php` action
* All actions are updated to remove node's version warnings
* Allow `composer test` script to take extra option (for `phpunit`)

Depend on https://github.com/pact-foundation/pact-php/pull/495

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,11 +54,6 @@ jobs:
     name: PHP ${{ matrix.php }} on ${{ matrix.operating-system }} with ${{ matrix.dependencies }} dependencies
 
     steps:
-      - name: Tune Windows Network
-        if: ${{ runner.os == 'Windows' }}
-        shell: pwsh
-        run: Disable-NetAdapterChecksumOffload -Name * -TcpIPv4 -UdpIPv4 -TcpIPv6 -UdpIPv6
-
       - uses: actions/checkout@v3
         name: Checkout repository
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,7 @@
-name: Pact-PHP Code Analysis & Test
+name: Code Analysis & Test
 
 on:
-  push: 
+  push:
   pull_request:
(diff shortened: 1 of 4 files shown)


Lewiscowles1986:
I've raised an issue against the setup-php action.

PR #503, opened 2024-03-02
feat: Allow set interaction key for v4

tienvx:
* Allow setting interaction/message interaction's key 
* This is for v4, using this feature on <= v3 doesn't have any effect to the pact file
* Update compatibility suite to use this feature

I'm not sure:
  * Did other languages implement it for end-user?
  * How did other languages implement for end-user?
  * Should all languages follow the same naming convention?

So I invented 3 methods in builder:

```php
 $config = new MockServerConfig();
$config->setPactSpecificationVersion('4.0.0');
$builder = new InteractionBuilder($config);
$builder
    ->key('key')
    ->pending(true) // other PR
    ->comments(['key' => 'value']) // other PR
    ->given('Get Goodbye')
    ->uponReceiving('A get request to /goodbye/{name}')
    ->with($request)
    ->willRespondWith($response);
```

Is this okay?

Should I change to `setKey`, `setPending` `setComments`?
Should I change to `withKey`, `withPending`, `withComments` ?

What do you think? cc @YOU54F @JP-Ellis

### DIFF
--- a/compatibility-suite/suites/v4/message/provider.yml
+++ b/compatibility-suite/suites/v4/message/provider.yml
@@ -11,6 +11,7 @@ default:
                     - '@interactions_storage'
                     - '@message_pact_writer'
                     - '@provider_verifier'
+                    - '@parser'
 
             filters:
                 tags: "@provider&&@message"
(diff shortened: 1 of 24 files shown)


Lewiscowles1986 (approved):

JP-Ellis:
I think the convention of prefixing methods with `set` or `with` or nothing is really up to you and what makes most sense in the language.

For example, languages that allow overloading of functions might have the same function as both the setter and getter based on the presence of an argument. Whereas languages which do not support this might instead have separate functions. You should follow what makes most sense for PHP.

In Python, I'm also  considering having a clear nomenclature to distinguish whether the pact is being modified or the specific interaction. I haven't settled yet on the nomenclature yet though.

PR #463, opened 2024-02-06
test(compatibility-suite): Fix compatibility suite v1 hang

tienvx:
I think recent update to pact broker `2.115` or `2.116` cause pact broker to hang. Move to `2.114` a month ago solve the issue

PR #514, opened 2024-03-05
Revert "test(compatibility-suite): Fix compatibility suite v1 hang"

YOU54F:
Reverts pact-foundation/pact-php#463


tienvx:
Do you want to revert so that it can catch this problem when it happen again? If so I will approve.

YOU54F:
yes please! it should have been caught by the test script in the repo but wasn't using the correct tag due to a sed error so it was testing the released version and not the to be released version. 

both this suite and pact pythons suites caught this so thank you!

PR #516, opened 2024-03-06
test: Test Verifier

tienvx:
* Remove `return $this`
* Use mocks in `VerifierTest`


Lewiscowles1986:
This is quite the change to usage to go from `self` to `void`, especially with a builder pattern.

tienvx:
They are all `private` methods. They are not belong to builder pattern. They can be called in any order (except `newHandle` need to be called first to get the handle of course)

Lewiscowles1986 (approved):

PR #517, opened 2024-03-06
fix: Fix 'call to unknown method' warning in ide

tienvx:
```php
interface UrlInterface
{
    public function setUrl(UriInterface $url): self;

    public function setToken(?string $token): self;

    public function setUsername(string $username): self;

    public function setPassword(string $password): self;
}

interface BrokerInterface extends UrlInterface
{
    // ...
}


        $broker = (new Broker())
            ->setUrl($url)
            ->setToken($token)
            ->setUsername($username)
            ->setPassword($password)
            ->setEnablePending($enablePending) // ide warning here
            ->setIncludeWipPactSince($wipPactSince)
            ->setProviderTags($providerTags)
            ->setProviderBranch($providerBranch)
            ->setConsumerVersionSelectors($consumerVersionSelectors)
            ->setConsumerVersionTags($consumerVersionTags);
```

### DIFF
--- a/src/PhpPact/Standalone/ProviderVerifier/Model/Source/Url.php
+++ b/src/PhpPact/Standalone/ProviderVerifier/Model/Source/Url.php
@@ -16,7 +16,7 @@ class Url implements UrlInterface
         return $this->url;
     }
 
-    public function setUrl(UriInterface $url): self
+    public function setUrl(UriInterface $url): static
     {
         $this->url = $url;
 
@@ -28,7 +28,7 @@ class Url implements UrlInterface
         return $this->token;
     }
 
-    public function setToken(?string $token): self
+    public function setToken(?string $token): static
     {
         $this->token = $token;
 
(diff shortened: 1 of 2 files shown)


Lewiscowles1986:
The issue with this is that it's not static. Is this IDE specific, and if so, would some IDE-specific annotation or work-around not work better for this?

In this case, self seems like the right type, but thinking very hard, perhaps static means something else as a return type.

tienvx:
> Is this IDE specific

Yes

> In this case, self seems like the right type, but thinking very hard, perhaps static means something else as a return type. Late-static bound perhaps?

I simply use this feature from PHP `8.0` https://php.watch/versions/8.0/static-return-type

> would some IDE-specific annotation or work-around not work better for this?

According to the link above, I think `@return static` also can do it.

Lewiscowles1986 (approved):

tienvx:
> returning UrlInterface will also work without static for this.

I replaced `self` by `UrlInterface` and it doesn't work.

Any way, I will close this PR for now and focus on writing more tests which is more important.

PR #527, opened 2024-03-08
test: Test ArrayData::createFrom() with empty array

tienvx:


tests/PhpPact/FFI/Model/ArrayDataTest.php
```diff
@@ -8,6 +8,11 @@
 
 class ArrayDataTest extends TestCase
 {
+    public function testCreateFromEmptyArray(): void
+    {
+        $this->assertNull(ArrayData::createFrom([]));
+    }
```
Lewiscowles1986, line 14: So Empty array leads to null, but what is the follow-up of that behaviour?

tienvx: The ffi method will receive `null` for both `items_pointer` and `items_length`. For example:

```php
$this->client->call(
    'pactffi_verifier_set_consumer_filters',
    $this->handle,
    null, // consumer_filters
    null // consumer_filters_len
);
```

Then rust will receive empty `consumers` vector.

Lewiscowles1986: Can we go one level deeper / higher-level, and discuss the end user / developer outcome?

tienvx: I will create a PR to test this behavior

tienvx: Created a PR to test this behavior https://github.com/pact-foundation/pact-php/pull/537

Lewiscowles1986: This one is 5 lines, lets pull this into 537. It's still not clear what it is doing, which is what my comment was about, but at least together there is a larger picture.

tienvx: merged into 537

tienvx: Here are some additional code:

When array is not empty:

```php
$providerTags = ['tag1', 'tag2', 'tag3'];
$arrayData = ArrayData::createFrom($providerTags);
$this->client->call(
    'pactffi_verifier_set_publish_options',
    $this->handle,
    ...
    $arrayData->getItems(), // FFI\CData
    $arrayData->getSize(), // int
    ....
);
```

When array is empty:

Because PHP doesn't allow empty array to be converted to FFI\CData. There is an error if we try to do so:


```php
FFI::new("char*[0]");

PHP Fatal error:  Uncaught FFI\Exception: Cannot instantiate FFI\CData of zero size
```

That's why `ArrayData::createFrom([])` need to return null. Rust will check if pointer is null then convert it into empty vector.

```php
$providerTags = [];
$nullValue = ArrayData::createFrom($providerTags);
$this->client->call(
    'pactffi_verifier_set_publish_options',
    $this->handle,
    ...
    null,
    null
    ....
);
```

PR #528, opened 2024-03-08
test: Test FFI Client

tienvx:


Lewiscowles1986 (commented):
These feel like useless changes. Would you mind explaining a little bit as to the value of this code?

tienvx:
The purpose of this test is only to increase code coverage

Lewiscowles1986:
Right but in this case it's nonsense coverage. If you were to run infection over this type of coverage it might still be low, because as one of the tests states, there are no assertions. We're simply expecting that calling something won't explode.

tienvx:
Replaced `pactffi_string_delete` by `pactffi_check_regex` which return a result. Added an assertion to assert the result. Does it look a bit better?

Lewiscowles1986 (approved):

PR #529, opened 2024-03-08
test: Test GrpcMockServer

tienvx:
* Reuse `ClientTrait`
* Add `GrpcMockServerTest`

### DIFF
--- a/tests/PhpPact/Consumer/Service/MockServerTest.php
+++ b/tests/PhpPact/Consumer/Service/MockServerTest.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace PhpPactTest\Consumer\Driver\Pact;
+namespace PhpPactTest\Consumer\Service;
 
 use FFI;
 use PhpPact\Config\PactConfigInterface;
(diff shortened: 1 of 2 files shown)


tests/PhpPact/Consumer/Service/MockServerTest.php
```diff
@@ -141,17 +143,8 @@ public function testCleanUp(): void
...
-        $this->client
-            ->expects($this->exactly(count($calls)))
-            ->method('call')
-            ->willReturnCallback(function (...$args) use (&$calls) {
-                $call = array_shift($calls);
-                $return = array_pop($call);
-                $this->assertSame($call, $args);
-
-                return $return;
-            });
```
Lewiscowles1986, line 155: so is this assertion now in the trait, simplifying responsibility of getTransport?

tests/PhpPact/Plugins/Protobuf/Service/GrpcMockServerTest.php
```diff
@@ -0,0 +1,20 @@
...
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->mockServer = new GrpcMockServer($this->client, $this->pactDriver, $this->config);
+    }
+
+    protected function getTransport(bool $secure): string
+    {
+        return 'grpc';
+    }
```
Lewiscowles1986, line 19: I'm confused by this as well. I Thought gRPC had to use https and HTTP2

tests/PhpPact/Consumer/Service/MockServerTest.php
tienvx: There are 2 changes that are not related to each other:
* Remove `assertClientCalls()`
* Add `getTransport()`

Because these 2 changes happen at the same time, at the same place, then git think that I replace `assertClientCalls()` by `getTransport()`, but I didn't. I removed `assertClientCalls()` first, then I add `getTransport()`, which is not related to the other one.

Probably I should add `getTransport()` to other position to make it clearer

tienvx: I think I understand your point: You want to split `assertClientCalls()` removing change into another PR?

tests/PhpPact/Plugins/Protobuf/Service/GrpcMockServerTest.php
tienvx: I don't understand much about the pact's plugin system. Here is what I "know":

`Transports` is one of a few **new terms** defined by Pact core when implementing [plugins](https://docs.pact.io/plugins/quick_start#introduction-to-plugins).

There are 3 transports so far that I know:
* core/transport/http
* core/transport/https
* plugin/protobuf/transport/grpc

All I know is: when I need to use grpc mock server, I need to use `grpc` transport provided by protobuf plugin, not `http` or `https` provided by the core.

tests/PhpPact/Consumer/Service/MockServerTest.php
tienvx: Will add this change back in another PR

Lewiscowles1986 (approved):

PR #537, opened 2024-03-09
test: Test verifier with empty array config

tienvx:
This PR test behavior when some verifier's config is empty array.

https://github.com/pact-foundation/pact-php/pull/527 is now part of this PR

### DIFF
--- a/tests/PhpPact/Standalone/ProviderVerifier/VerifierTest.php
+++ b/tests/PhpPact/Standalone/ProviderVerifier/VerifierTest.php
@@ -37,6 +37,10 @@ class VerifierTest extends TestCase
         $this->logger = $this->createMock(LoggerInterface::class);
         $this->client = $this->createMock(ClientInterface::class);
         $this->handle = FFI::new('int');
+    }
+
+    private function setUpCalls(bool $hasProviderTags = true, bool $hasFilterConsumerNames = true): void
+    {
         $this->config->getCallingApp()
             ->setName($callingAppName = 'calling app name')
             ->setVersion($callingAppVersion = '1.2.3');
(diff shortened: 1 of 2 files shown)


tests/PhpPact/Standalone/ProviderVerifier/VerifierTest.php
```diff
@@ -80,20 +84,41 @@ protected function setUp(): void
...
             ['pactffi_verifier_set_verification_options', $this->handle, $disableSslVerification, $requestTimeout, null],
-            ['pactffi_verifier_set_publish_options', $this->handle, $providerVersion, $buildUrl, $this->isInstanceOf(CData::class), count($providerTags), $providerBranch, null],
-            ['pactffi_verifier_set_consumer_filters', $this->handle, $this->isInstanceOf(CData::class), count($filterConsumerNames), null],
+            [
+                'pactffi_verifier_set_publish_options',
+                $this->handle,
+                $providerVersion,
+                $buildUrl,
+                $hasProviderTags ? $this->isInstanceOf(CData::class) : null,
+                $hasProviderTags ? count($providerTags) : null,
```
JP-Ellis, line 93: Just double checking, is the default value meant to be `null`, or should it be an empty array? The latter makes more sense to me.

tienvx: The default value is `null`, not empty array. Test will fail with this error if I change to empty array:


```
1) PhpPactTest\Standalone\ProviderVerifier\VerifierTest::testConstruct with data set #1 (false, false)
Failed asserting that null is identical to Array &0 [].
```

I explained why the value is `null` in this comment https://github.com/pact-foundation/pact-php/pull/527#discussion_r1518582953

JP-Ellis: Thanks! After reading the comment, makes sense :)

PR #552, opened 2024-04-18
chore: arrayContains and eachValue example for query and header

tienvx:
Depend on https://github.com/pact-foundation/pact-php/pull/551

### DIFF
--- a/example/matchers/consumer/tests/Service/MatchersTest.php
+++ b/example/matchers/consumer/tests/Service/MatchersTest.php
@@ -27,17 +27,26 @@ class MatchersTest extends TestCase
             ->setMethod('GET')
             ->setPath($this->matcher->regex('/matchers', '^\/matchers$'))
             ->setQuery([
-                'pages' => $this->matcher->regex([1, 22], '\d+'), // arrayContains, eachKey, eachValue matchers are not working with query
-                'locales[]' => $this->matcher->regex(['en-US', 'en-AU'], '^[a-z]{2}-[A-Z]{2}$'), // Use `locales[]` instead of `locales` syntax if provider use PHP language
+                'pages' => $this->matcher->eachValue([1, 22], [
+                    $this->matcher->regex(null, '\d+')
+                ]),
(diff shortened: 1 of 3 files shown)


example/matchers/consumer/tests/Service/MatchersTest.php
```diff
@@ -27,17 +27,26 @@ public function testGetMatchers(): void
...
-                'pages' => $this->matcher->regex([1, 22], '\d+'), // arrayContains, eachKey, eachValue matchers are not working with query
-                'locales[]' => $this->matcher->regex(['en-US', 'en-AU'], '^[a-z]{2}-[A-Z]{2}$'), // Use `locales[]` instead of `locales` syntax if provider use PHP language
+                'pages' => $this->matcher->eachValue([1, 22], [
+                    $this->matcher->regex(null, '\d+')
+                ]),
+                // 'pages' => $this->matcher->regex([1, 22], '\d+'), // Alternative approach
+                'locales[]' => $this->matcher->eachValue(['en-US', 'en-AU'], [
+                    $this->matcher->regex(null, '^[a-z]{2}-[A-Z]{2}$'),
+                ]),  // Use `locales[]` instead of `locales` syntax if provider use PHP language
+                // 'locales[]' => $this->matcher->regex(['en-US', 'en-AU'], '^[a-z]{2}-[A-Z]{2}$'), // Alternative approach
```
JP-Ellis, line 37: Do we need the commented alternative approaches?

Given it is in the examples directory, it might be worth keeping for pedagogical purposes, but I personally think it would be more readable as its own large stand-alone comment which might explain the pros and cons of both alternatives.

JP-Ellis (approved):

example/matchers/consumer/tests/Service/MatchersTest.php
tienvx: I made some changes to make it more readable https://github.com/pact-foundation/pact-php/pull/554

PR #557, opened 2024-04-19
chore: Use both pactffi_set_comment and pactffi_add_text_comment

tienvx:
`pactffi_set_comment` does work, but there is a problem. `pactffi_add_text_comment` is a better replacement for it.

See these links for more information:
* https://github.com/pact-foundation/pact-compatibility-suite/pull/8
* https://github.com/pact-foundation/pact-specification/issues/45
* https://github.com/pact-foundation/pact-reference/pull/401

### DIFF
--- a/compatibility-suite/tests/Context/V4/Http/ConsumerContext.php
+++ b/compatibility-suite/tests/Context/V4/Http/ConsumerContext.php
@@ -77,7 +77,7 @@ final class ConsumerContext implements Context
      */
     public function aCommentIsAddedToTheHttpInteraction(string $value): void
     {
-        $this->interaction->addComment($value);
+        $this->interaction->addTextComment($value);
     }
 
     /**
--- a/compatibility-suite/tests/Context/V4/Http/ConsumerContext.php
+++ b/compatibility-suite/tests/Context/V4/Http/ConsumerContext.php
@@ -77,7 +77,7 @@ final class ConsumerContext implements Context
      */
     public function aCommentIsAddedToTheHttpInteraction(string $value): void
     {
-        $this->interaction->setComments(['text' => json_encode([$value])]);
+        $this->interaction->addComment($value);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


JP-Ellis:
I think it might be best to extend the commenting functionality, instead of just replacing the current method. I did something similar in Python by [adding two methods](https://github.com/pact-foundation/pact-python/blob/1ee19a97bdefe4f449967539468a86b3598b0e09/src/pact/v3/interaction/_base.py#L341-L388):

```python
    def set_comment(self, key: str, value: Any | None) -> Self:  # noqa: ANN401
        """
        Set a comment for the interaction.

        This is used by V4 interactions to set a comment for the interaction. A
        comment consists of a key-value pair, where the key is a string and the
        value is anything that can be encoded as JSON.

        Args:
            key:
                Key for the comment.

            value:
                Value for the comment. This must be encodable using
                [`json.dumps(...)`][json.dumps], or an existing JSON string. The
                value of `None` will remove the comment with the given key.

        # Warning

        This function will overwrite any existing comment with the same key. In
        particular, the `text` key is used by `add_text_comment`.
        """
        if isinstance(value, str) or value is None:
            pact.v3.ffi.set_comment(self._handle, key, value)
        else:
            pact.v3.ffi.set_comment(self._handle, key, json.dumps(value))
        return self

    def add_text_comment(self, comment: str) -> Self:
        """
        Add a text comment for the interaction.

        This is used by V4 interactions to set arbitrary text comments for the
        interaction.

        Args:
            comment:
                Text of the comment.

        # Warning

        Internally, the comments are appended to an array under the `text`
        comment key. Care should be taken to ensure that conflicts are not
        introduced by
        [`set_comment`][pact.v3.interaction.Interaction.set_comment].
        """
        pact.v3.ffi.add_text_comment(self._handle, comment)
        return self
```

This allows end-users to add comments as per the original intentions behind the capability, while also allowing end-users to make use of the full capabilities of the Pact specification.

I would make sure to be really clear in the documentation of the possible way they can interact to hopefully avoid any unexpected side-effects.

tienvx:
Added method `pactffi_set_comment` back

JP-Ellis (approved):

PR #559, opened 2024-04-22
docs: Reorganize provider usage in readme

tienvx:
* Reorganize 'provider usage' section in readme
* Update table of contents
* Use recommended consumer version selectors (`deployedOrReleased` and `mainBranch` instead of `tag` and `latest`) https://docs.pact.io/pact_broker/advanced_topics/consumer_version_selectors#recommended

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,31 +269,50 @@ All of the following code will be used exclusively for Providers. This will run 
 Create a single unit test function. This will test all defined consumers of the service.
 
 ```php
+protected function setUp(): void
+{
+    // Start API
+}
+
+protected function tearDown(): void
+{
+    // Stop API
+}
+
 public function testPactVerifyConsumers(): void
 {
     $config = new VerifierConfig();
-    $config
-        ->setProviderName('someProvider') // Providers name to fetch.
-        ->setProviderVersion('1.0.0') // Providers version.
-        ->setProviderTags('prod' ,'dev')
-        ->setProviderBranch('main')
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -266,52 +264,57 @@ See how to use at https://github.com/pactflow/actions/tree/main/publish-pact-fil
...
+        ->setBuildUrl(new Uri('http://build.domain.com'))
+        ->setFilterConsumerNames('someConsumer', 'otherConsumer')
+        ->setFilterDescription('Send POST to create')
+        ->setFilterNoState(true)
+        ->setFilterState('state')
+        ->setPublishResults(true)
+        ->setDisableSslVerification(true)
+        ->setStateChangeAsBody(false)
+        ->setStateChangeTeardown(true)
+        ->setRequestTimeout(500);
```
JP-Ellis, line 294: I think that for the front-page README, this is quite verbose and might overwhelm some people.

If a lot of the options have sane defaults, I would simply ignore them and focus on the important configuration options only. If they need these additional configuration options, they can search the more in-depth docs.

JP-Ellis (changes requested):

README.md
tienvx: So I think I will try to use configurations from my personal projects.

tienvx: I updated the example code, copy from `UPGRADE-10.0.md` and my personal project

JP-Ellis (approved):

PR #568, opened 2024-04-25
chore: Fix bool to int casting errors

tienvx:
Fix errors like:

```
Parameter #1 $count of method PHPUnit\Framework\TestCase::exactly() expects int, bool given.
```

in https://github.com/pact-foundation/pact-php/pull/564

### DIFF
--- a/tests/PhpPact/Consumer/Driver/Interaction/InteractionDriverTest.php
+++ b/tests/PhpPact/Consumer/Driver/Interaction/InteractionDriverTest.php
@@ -101,7 +101,7 @@ class InteractionDriverTest extends TestCase
         ];
         $this->assertClientCalls($calls);
         $this->mockServer
-            ->expects($this->exactly($startMockServer))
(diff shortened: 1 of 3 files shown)


JP-Ellis:
I feel like this may not be the best way to resolve this. Is it possible to expand the definition of `expects` to allow Booleans? Or use a different method?

tienvx:
I have another approach:

```php
$this->exactly($boolValue ? 1 : 0)
```

It's shorter. How about it?

tienvx:
> Is it possible to expand the definition of expects to allow Booleans?

I don't think it's possible. The first parameter of the method `expects`  need to be instance of `PHPUnit\Framework\MockObject\Rule\InvocationOrder` type.

JP-Ellis:
It's all good, revisiting it, I think your initial solution is fine.

I think it's just coming from another other testing framework, I had something else in mind (which probably wouldn't apply to PHP).

JP-Ellis (approved):

PR #580, opened 2024-04-26
chore: PHPStan ignore line

tienvx:
Ignore this error:

```
Cannot access offset 0|1|2|3 on FFI\CData.
```

For https://github.com/pact-foundation/pact-php/pull/564


JP-Ellis (approved):
Approving, but just wanting to check: Is it possible to ignore a specific error?

By analogy, in Python we have:

```python
x: int = "foo" # noqa   <-- Generic 'ignore all errors'; bad practice
x: int = "foo" # noqa: some-error <-- Only ignores 'some-error'; best practice
```

tienvx:
> Is it possible to ignore a specific error?

Yes, it's possible, but only when PHPStan version `1.11` is released (sometime in May). The current version is `1.10.67`, which is the last version of `1.10`.

So I use 2 solutions, and will remove the other when `1.11` is released.

PR #584, opened 2024-04-26
docs(ci): add trigger update to docs.pact.io

YOU54F:
sync job not currently present in docs.pact.io repo on master, but will be later today

### DIFF
--- a/.github/workflows/trigger_pact_docs_update.yml
+++ b/.github/workflows/trigger_pact_docs_update.yml
@@ -1,0 +1,22 @@
+name: Trigger update to docs.pact.io
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    paths:
+      - '**.md'
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger docs.pact.io update workflow
+        run: |
+          curl -X POST https://api.github.com/repos/pact-foundation/docs.pact.io/dispatches \
+                -H 'Accept: application/vnd.github.everest-preview+json' \
+                -H "Authorization: Bearer $GITHUB_TOKEN" \
+                -d '{"event_type": "pact-php-docs-updated"}'
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger_pact_docs_update.yml
```diff
@@ -0,0 +1,22 @@
...
+      - '**.md'
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger docs.pact.io update workflow
+        run: |
+          curl -X POST https://api.github.com/repos/pact-foundation/docs.pact.io/dispatches \
+                -H 'Accept: application/vnd.github.everest-preview+json' \
```
tienvx, line 18: I search for the keyword "application/vnd.github.everest-preview+json" and I think this is quite old. Do we aware about this? https://developer.github.com/changes/3/

Maybe this is new syntax https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#create-a-repository-dispatch-event

.github/workflows/trigger_pact_docs_update.yml
```diff
@@ -0,0 +1,22 @@
...
+                -H "Authorization: Bearer $GITHUB_TOKEN" \
+                -d '{"event_type": "pact-php-docs-updated"}'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GHTOKENFORTRIGGERINGPACTDOCSUPDATE }}
```
tienvx, line 22: It's hard to read. If we are already using it for several projects I fine with it.

tienvx (approved):
Approved. There are some comments but if we are already using this code several places it's fine to me.

.github/workflows/trigger_pact_docs_update.yml
YOU54F: it's set at an organisation level, but agree it is hard to read.

If we update the webhook, I'm happy to change this to be easier to grok (as we will have to update each repos workflow anyway)

PR #589, opened 2024-05-02
refactor: Apply Visitor pattern for MatchingField matcher

tienvx:
There are a lot of `if` conditions in `PluginFormatter`. I think `Visitor` pattern can be applied to solve this.

### DIFF
--- a/src/PhpPact/Consumer/Matcher/Formatters/PluginFormatter.php
+++ b/src/PhpPact/Consumer/Matcher/Formatters/PluginFormatter.php
@@ -28,9 +28,6 @@ class PluginFormatter implements FormatterInterface
             throw new GeneratorNotRequiredException('Generator is not support in plugin');
         }
 
-        if ($matcher instanceof MatchingField) {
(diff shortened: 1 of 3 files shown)


src/PhpPact/Consumer/Matcher/Formatters/PluginFormatter.php
```diff
@@ -51,7 +48,7 @@ public function format(MatcherInterface $matcher): string
         throw new MatcherNotSupportedException(sprintf("Matcher '%s' is not supported by plugin", $matcher->getType()));
```
tienvx, line 48: I think I will replace this exception by `BadMethodCallException('Method not allowed')` once I removed all of the `if` conditions.

JP-Ellis (approved):

src/PhpPact/Consumer/Matcher/Formatters/PluginFormatter.php
tienvx: I was wrong. This method is still required for rules for EachValue/EachKey matchers.

PR #591, opened 2024-05-02
refactor: Apply Visitor pattern for EachValue/EachKey matchers

tienvx:
There are a lot of `if` conditions in `PluginFormatter.` I think `Visitor` pattern can be applied to solve this.


tests/PhpPact/Consumer/Matcher/Formatters/PluginFormatterTest.php
```diff
@@ -79,6 +83,8 @@ public function testInvalidValue(MatcherInterface $matcher, string $type): void
 
     #[TestWith([new MatchingField('product')])]
     #[TestWith([new NotEmpty('test')])]
+    #[TestWith([new EachKey(["doesn't matter"], [])])]
+    #[TestWith([new EachValue(["doesn't matter"], [])])]
```
tienvx, line 87: Will revert in https://github.com/pact-foundation/pact-php/pull/593

PR #592, opened 2024-05-02
refactor: Apply Visitor pattern for ContentType matcher

tienvx:
There are a lot of`if` conditions in `PluginFormatter`. I think `Visitor` pattern can be applied to solve this.


tests/PhpPact/Consumer/Matcher/Formatters/PluginFormatterTest.php
```diff
@@ -79,6 +79,7 @@ public function testInvalidValue(MatcherInterface $matcher, string $type): void
 
     #[TestWith([new MatchingField('product')])]
     #[TestWith([new NotEmpty('test')])]
+    #[TestWith([new ContentType('application/xml')])]
```
tienvx, line 82: Will revert in https://github.com/pact-foundation/pact-php/pull/593

PR #597, opened 2024-05-03
feat: Allow combining matchers

tienvx:
Implemented this feature https://github.com/pact-foundation/pact-reference/pull/405

### DIFF
--- a/example/matchers/consumer/tests/Service/MatchersTest.php
+++ b/example/matchers/consumer/tests/Service/MatchersTest.php
@@ -113,6 +113,24 @@ class MatchersTest extends TestCase
                     [$this->matcher->regex(null, 'car|bike|motorbike')]
                 ),
                 'url' => $this->matcher->url('http://localhost:8080/users/1234/posts/latest', '.*(\\/users\\/\\d+\\/posts\\/latest)$', false),
+                'matchAll' => $this->matcher->matchAll(
+                    ['desktop' => '2000 usd'],
+                    [
+                        $this->matcher->eachKey(
+                            ['laptop' => '1500 usd'],
+                            [$this->matcher->regex(null, 'laptop|desktop|mobile|tablet')]
+                        ),
+                        $this->matcher->eachValue(
+                            ['mobile' => '500 usd'],
+                            [
+                                $this->matcher->includes('usd'),
+                                $this->matcher->regex(null, '\d+ \w{3}')
+                            ],
+                        ),
+                        $this->matcher->atLeast(2),
+                        $this->matcher->atMost(3),
+                    ],
+                ),
 
                 // Don't mind this. This is for demonstrating what query values provider will received.
                 'query' => [
@@ -199,6 +217,7 @@ class MatchersTest extends TestCase
                 'vehicle 1' => 'car',
             ],
             'url' => 'http://localhost:8080/users/1234/posts/latest',
+            'matchAll' => ['desktop' => '2000 usd'],
 
             // Don't mind this. This is for demonstrating what query values provider will received.
             'query' => [
--- a/src/PhpPact/Consumer/Matcher/Formatters/PluginFormatter.php
+++ b/src/PhpPact/Consumer/Matcher/Formatters/PluginFormatter.php
@@ -3,13 +3,17 @@
 namespace PhpPact\Consumer\Matcher\Formatters;
 
 use PhpPact\Consumer\Matcher\Exception\GeneratorNotRequiredException;
+use PhpPact\Consumer\Matcher\Exception\InvalidValueException;
 use PhpPact\Consumer\Matcher\Exception\MatcherNotSupportedException;
 use PhpPact\Consumer\Matcher\Exception\MatchingExpressionException;
 use PhpPact\Consumer\Matcher\Matchers\AbstractDateTime;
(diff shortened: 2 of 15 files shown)


mefellows:
Nice! Wow you got onto this feature quickly, I need to get my head into how a consumer might look, ad perhaps try with Pact JS/Go to test it out.

Would a review early next week work or is that going to be too late for you?

src/PhpPact/Consumer/Matcher/Formatters/PluginFormatter.php
```diff
@@ -40,13 +44,22 @@ public function format(MatcherInterface $matcher): string
...
 
         if (in_array($matcher->getType(), self::MATCHERS_WITHOUT_CONFIG)) {
             return $this->formatMatchersWithoutConfig($matcher);
         }
         if ($matcher instanceof AbstractDateTime || $matcher instanceof Regex || $matcher instanceof ContentType) {
             return $this->formatMatchersWithConfig($matcher);
         }
+        if ($matcher instanceof MatchAll) {
+            return $this->formatMatchAllMatchers($matcher);
+        }
```
tienvx, line 62: Too much if conditions. I tried to solve this code smell by applying `Visitor` design pattern, but end up reverting the PRs, because there is a problem I can't solve it right now. I guess we will need to solve this one later.

tienvx:
A review early next week would be great. I will focus on other things for now. Thanks

src/PhpPact/Consumer/Matcher/Formatters/PluginFormatter.php
tienvx: Refactored in https://github.com/pact-foundation/pact-php/pull/599.

Instead of applying `Visitor`, I use `Factory Method` (combined with `Bridge`) patterns.

This time, I implemented full code before sending PR to prevent reverting the PRs that are not working.

That's a very big PR. I am not patient so I send it to review without breaking down to smaller PRs. I think it will take 1,5 weeks to review all of them. I'm sorry but I don't have time right now.

YOU54F:
This looks great.

With regard to

> Pact JVM supports combining matching rules using the AND logical expression.

I was just looking to how this is implemented

https://github.com/pact-foundation/pact-jvm/blob/master/consumer/junit/README.md#combining-matching-rules-with-andor

It looks to support `AND` and `OR` operators. 

and with regard to naming of `MatchAll`, if we were to have the ability to do an `OR` operation, what would that be called.

ahhh looking further, it doesn't look like `OR` is supported as part of https://github.com/pact-foundation/pact-reference/pull/405

I just took a look at the matchers in pact-js and pact-go, to think about how the DSL would look. ie would be want to introduce a `MatchAll` function there. The pact-jvm DSL is clunkier for matcher bodies than other languages, so your approach might be preferable.

- [pact-js docs](https://github.com/pact-foundation/pact-js/blob/master/docs/matching.md)
- [pact-js v3 matchers code](https://github.com/pact-foundation/pact-js/blob/master/src/v3/matchers.ts)
- [pact-go matchers code](https://github.com/pact-foundation/pact-go/blob/master/matchers/matcher_v3.go)

tienvx:
> ahhh looking further, it doesn't look like `OR` is supported

Yes, it doesn't support. But I assume it will support in the future, and I suggest `matchAny` for it.

example/matchers/consumer/tests/Service/MatchersTest.php
```diff
@@ -113,6 +113,24 @@ public function testGetMatchers(): void
                     [$this->matcher->regex(null, 'car|bike|motorbike')]
                 ),
                 'url' => $this->matcher->url('http://localhost:8080/users/1234/posts/latest', '.*(\\/users\\/\\d+\\/posts\\/latest)$', false),
+                'matchAll' => $this->matcher->matchAll(
```
YOU54F, line 116: Maybe I am having a dim moment, but if `like` expects all values to match, and can accept nested matchers, why do we need to expose a new matcher to the end user.

example/matchers/consumer/tests/Service/MatchersTest.php
```diff
@@ -113,6 +113,24 @@ public function testGetMatchers(): void
...
+                    ['desktop' => '2000 usd'],
```
YOU54F, line 117: is this the generator?

should it be enforced to have `atLeast(2)` in length

it gets stored in the body of the pact as

```json
            "matchAll": {
              "desktop": "2000 usd"
            },
```

In our incoming query we send 

`['desktop' => '2000 usd']` (line 220)

Which doesn't match the `atLeast(2)` length restriction.

tienvx: > is this the generator?

No, it's example value.

> should it be enforced to have atLeast(2) in length

Because it's just example value, it doesn't have to. But yes, it make sense to be at least 2 items. I'm just lazy to do so.

> In our incoming query we send `['desktop' => '2000 usd']` (line 220)

It's response from mock server, not request. And mock server just return the example value, doesn't validate at this point.

tienvx:
Self-reviewed

example/matchers/consumer/tests/Service/MatchersTest.php
mefellows: Yes, that's an example of when you might like to use both.

PR #599, opened 2024-05-05
Formatter factory method

tienvx:
Depend on:
* https://github.com/pact-foundation/pact-php/pull/597
* https://github.com/pact-foundation/pact-php/pull/598

Apply `Bridge` and `Factory Method` design patterns to refactor formatters

Here is the full diagram:

matchers-formatters


example/csv/consumer/tests/Service/HttpClientServiceTest.php
```diff
@@ -18,7 +17,7 @@ class HttpClientServiceTest extends TestCase
 {
     public function testGetCsvFile(): void
     {
-        $matcher = new Matcher(new PluginFormatter());
```
YOU54F, line 21: This looks like a nicer experience for end users (removing an additional import)

YOU54F:
Nice change! It's a biggie so appreciate you splitting out the two PR's prior, as it helped me focus on them.

I'm looking at the end user interface changes in the examples, and that looks make it easier (removing an import and having as a option), considering your touched ~140 files, have minimal changes in your tests and everything green makes me a very confident person.

Appreciate the diagram!

So in a nutshell does this make it easier to maintainers/contributors to maintain/extend these interfaces in future? Whats the big selling point of moving to the new pattern.

Looks good to me though, haven't ran it locally but he have 💚 ci

example/csv/consumer/tests/Service/HttpClientServiceTest.php
tienvx: :hugs: 

There is a small problem with these code: user need to figure out if a plugin is being used or not.

tienvx:
> So in a nutshell does this make it easier to maintainers/contributors to maintain/extend these interfaces in future?

Yes, you are correct. This is the whole point of this PR.

tienvx:
Self-reviewed

PR #601, opened 2024-05-12
refactor: Add formatters for Min/Max Type matchers

tienvx:
* Value for MinType/MaxType/MinMaxType matchers now is not array
* Add expression/json formatters for MinType/MaxType/MinMaxType matchers
* Remove ability to set number of examples. I don't think it's important.
  * Pact-JS has this ability
  * Pact-GO doesn't have this ability (https://github.com/pact-foundation/pact-go/blob/master/matchers/matcher_v3.go#L190-L193)
  * Pact-PHP is mixed . 1 matcher has (`constrainedArrayLike`). others don't
    * Now it's just like Pact-GO


tienvx:
Self-reviewed

PR #602, opened 2024-05-12
test: Test map matching for protobuf

tienvx:


tienvx:
Related issue https://github.com/pactflow/pact-protobuf-plugin/issues/62

YOU54F (approved):
good example @tienvx

PR #613, opened 2024-05-14
feat: Support Alpine

tienvx:
Thanks @YOU54F for the implementation. Here is the original code https://github.com/pact-foundation/pact-php/compare/master...YOU54F:pact-php:tx-support-alpine

We will need to update the Alpine ci once Github Actions Linux ARM is available.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -173,3 +173,37 @@ jobs:
       uses: coverallsapp/github-action@v2
       with:
         parallel-finished: true
+
+  test-alpine:
+    runs-on: ubuntu-latest
+    needs:
+      - php-cs
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [ amd64, arm64 ]
+
+    steps:
+      - uses: actions/checkout@v4
+        name: Checkout repository
+      - name: Set up QEMU
+        if: matrix.arch == 'arm64'
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
+        if: matrix.arch == 'arm64'
+        uses: docker/setup-buildx-action@v3
+      # - name: Build and cache
+      #   uses: docker/build-push-action@v5
+      #   with:
+      #     context: .
+      #     push: false
+      #     tags: "alpine-base:${{ matrix.arch }}"
+      #     cache-from: type=gha
+      #     cache-to: type=gha,mode=max
+      #     platforms: linux/${{ matrix.arch }}
+      - name: run musl # If shared-mime-info not installed - Expected binary contents to have content type 'image/jpeg' but detected contents was 'application/octet-stream'
+        if: matrix.arch == 'amd64'
+        run: docker run --platform=linux/${{ matrix.arch }} --rm -v $PWD:/home you54f/pact-php:alpine-base-${{ matrix.arch }} /bin/sh -c 'apk add --no-cache shared-mime-info && cd /home && composer install && composer gen-lib && composer test'
+      - name: run musl # If shared-mime-info installed - Expected binary contents to have content type 'image/jpeg' but detected contents was 'application/octet-stream'
+        if: matrix.arch == 'arm64'
+        run: docker run --platform=linux/${{ matrix.arch }} --rm -v $PWD:/home you54f/pact-php:alpine-base-${{ matrix.arch }} /bin/sh -c 'apk add --no-cache shared-mime-info && cd /home && composer install && composer gen-lib && composer test'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -192,18 +192,9 @@ jobs:
       - name: Set up Docker Buildx
         if: matrix.arch == 'arm64'
         uses: docker/setup-buildx-action@v3
-      # - name: Build and cache
-      #   uses: docker/build-push-action@v5
-      #   with:
-      #     context: .
-      #     push: false
-      #     tags: "alpine-base:${{ matrix.arch }}"
-      #     cache-from: type=gha
-      #     cache-to: type=gha,mode=max
-      #     platforms: linux/${{ matrix.arch }}
-      - name: run musl # If shared-mime-info not installed - Expected binary contents to have content type 'image/jpeg' but detected contents was 'application/octet-stream'
+      - name: run musl ${{ matrix.arch }} # If shared-mime-info not installed - Expected binary contents to have content type 'image/jpeg' but detected contents was 'application/octet-stream'
         if: matrix.arch == 'amd64'
-        run: docker run --platform=linux/${{ matrix.arch }} --rm -v $PWD:/home you54f/pact-php:alpine-base-${{ matrix.arch }} /bin/sh -c 'apk add --no-cache shared-mime-info && cd /home && composer install && composer gen-lib && composer test'
-      - name: run musl # If shared-mime-info installed - Expected binary contents to have content type 'image/jpeg' but detected contents was 'application/octet-stream'
+        run: docker run --env PACT_DO_NOT_TRACK=1 --platform=linux/${{ matrix.arch }} --rm -v $PWD:/home alpine:3.19 /bin/sh -c 'apk add --no-cache shared-mime-info php82-dev php82-ffi php82-pecl-grpc php82-sockets php82-tokenizer php82-dom php82-xml php82-xmlwriter php82-simplexml composer protoc protobuf-dev && cd /home && composer install && composer gen-lib && composer test'
+      - name: run musl ${{ matrix.arch }} # If shared-mime-info not installed - Expected binary contents to have content type 'image/jpeg' but detected contents was 'application/octet-stream'
         if: matrix.arch == 'arm64'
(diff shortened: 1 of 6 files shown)


example/protobuf-sync-message/provider/composer.json
```diff
@@ -16,11 +16,12 @@
             "rr": {
                 "version": "2023.3.10",
                 "variables": {
-                    "{$os}": "strtolower(PHP_OS_FAMILY)",
+                    "{$os}": "PHP_OS === 'Linux' && musl() === true ? 'unknown' : strtolower(PHP_OS_FAMILY)",
+                    "{$musl}": "PHP_OS === 'Linux' && musl() === true ? '-musl' : ''",
```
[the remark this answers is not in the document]
tienvx: (I think) I tried but I got this error:

```
FFI\Exception: Failed loading '/home/src/PhpPact/Standalone/Installer/Model/../../../../../bin/pact-ffi-lib/pact.so' (Error relocating /home/src/PhpPact/Standalone/Installer/Model/../../../../../bin/pact-ffi-lib/pact.so: __builtin_copysignq: symbol not found)
```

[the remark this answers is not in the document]
YOU54F: it's weird that the error is coming from the pact.so and not from the rr binary.

`pact-ffi-lib/pact.so: __builtin_copysignq: symbol not found)`

Maybe we just need to try in a vanilla alpine container with arm64, download the rr binary into it and exec it. 

If not, we could exclude the rr test on musl, until this PR is able to be delivered

https://github.com/roadrunner-server/roadrunner/issues/1875#issuecomment-1966394413

we could also offer a PR to build it under qemu with docker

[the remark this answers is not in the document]
tienvx: So probably I did something wrong there.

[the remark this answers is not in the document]
YOU54F: so the arm64 rr binary loads fine in an alpine container (need to find the right combo of extensions but there is actually an issue with loading the library on aarch64 musl) let me do some investigations and I'll get back to you on this.

Not sure of your hardware you have at home, but I have been testing aarch64 natively on apple silicon macbook m1 so have a bit of a quicker feedback loop.

Cheers for the efforts Tien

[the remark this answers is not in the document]
YOU54F: So looks to be an issue with `0.4.20` version of `libpact_ffi-linux-aarch64-musl.so` 
-  (https://github.com/pact-foundation/pact-reference/releases/download/libpact_ffi-v0.4.20/libpact_ffi-linux-aarch64-musl.so.gz)

Tested with 0.4.19, results in being able to load the shared library successfully.

## 0.4.20



Screenshot 2024-05-14 at 18 27 29

## 0.4.19

Screenshot 2024-05-14 at 18 29 01

I wonder if we need to pin our version of alpine to a earlier version, so that we are forward compat with as many alpine versions as possible for the shared lib.

https://github.com/pact-foundation/pact-reference/blob/e44458a8616856d3726470dd8f3884340999bf1e/rust/pact_ffi/release-linux.sh#L87-L102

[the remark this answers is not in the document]
YOU54F: Right got a passing run of musl-arm64 local and pushed up to gh actions and passing on there. used the pre-baked build for speed (the grpc compilation takes forever in the build step even locally)

https://github.com/pact-foundation/pact-php/compare/master...YOU54F:pact-php:tx-support-alpine

content-type hack needs the condition changes to mac arm64 or windows, based on the current testing.

still need to work out why the 0.4.20 aarch64 musl so isn't ok. 

We can build the so easily from the existing a file (which is working in pact-js-core PR) so I would probably start by building the so file in an older alpine container, and seeing how it fares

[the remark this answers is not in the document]
YOU54F: just to confirm, the none alpine variants of the road runner binary appear to work on alpine, so you can either restrict the condition to musl and x86_64, or just use the same binaries on musl/non musl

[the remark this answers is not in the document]
tienvx: I was a little bit confused on this one so I didn't follow you there.

README.md
```diff
@@ -270,6 +270,8 @@ _\*_ v3 support is limited to the subset of functionality required to enable lan
 | Linux   | arm64        | ✅         | 9.x +            |
 | Windows | x86_64       | ✅         | All              |
 | Windows | x86          | ✅         | 9.x -            |
+| Alpine  | x86_64       | ✅         | 10.x +           |
+| Alpine  | arm64        | ✅         | 10.x +           |
```
tienvx, line 274: I'm not sure about 9.x and below. I think Ruby standalone did support it, right?

YOU54F: Yes, it was supported with a workaround

for 9.x and below, users should be noted that they require extra dependencies, as the pact-ruby-standalone package is glibc based

https://github.com/pact-foundation/pact-ruby-standalone/wiki/Using-the-pact-ruby-standalone-with-Alpine-Linux-Docker

YOU54F: perfect thanks @tienvx

.github/workflows/build.yml
```diff
@@ -173,3 +173,37 @@ jobs:
...
+      #   with:
+      #     context: .
+      #     push: false
+      #     tags: "alpine-base:${{ matrix.arch }}"
+      #     cache-from: type=gha
+      #     cache-to: type=gha,mode=max
+      #     platforms: linux/${{ matrix.arch }}
+      - name: run musl # If shared-mime-info not installed - Expected binary contents to have content type 'image/jpeg' but detected contents was 'application/octet-stream'
+        if: matrix.arch == 'amd64'
+        run: docker run --platform=linux/${{ matrix.arch }} --rm -v $PWD:/home you54f/pact-php:alpine-base-${{ matrix.arch }} /bin/sh -c 'apk add --no-cache shared-mime-info && cd /home && composer install && composer gen-lib && composer test'
```
YOU54F, line 206: I can push these images to `pactfoundation` scope as an interim measure, otherwise I am happy to setup credentials in the repo to allow for publishing to DockerHub.

We could also publish to GitHub container registry, and then its nicely linked to the repo, and easy to setup creds. 

I am sure people would prefer having seen the images being published from a CI machine.

I noted there was an issue in loading the image from the cache in subsequent runs, which is why I ended up skipping the caching step

YOU54F: Infact I think we can get away with a vanilla image and install some requirements directly from apk rather than building grpc from source

```
docker run --platform=linux/${{ matrix.arch }} --rm -v $PWD:/home alpine:3.19 /bin/sh -c 'apk add --no-cache shared-mime-info php82-dev php82-ffi php82-pecl-grpc php82-sockets php82-tokenizer php82-dom php82-xml php82-xmlwriter composer protoc protobuf-dev && cd /home && composer install && composer gen-lib && composer test'
```

YOU54F: https://github.com/YOU54F/pact-php/blob/6d0e88b682377020e7b7b015fbe922ed744e4e58/.github/workflows/build.yml#L192

fixed up arm64 as it wasn't pulling the latest version of the csv plugin, so installing it manually in this run

https://github.com/pact-foundation/pact-plugins/issues/61

Need to update the manifest, and there wasn't a correct manifest file (which I've now added)

as a side note: does pact-php always try and grab the latest version of the plugin on the index, or can a user fix it to a particular version?

YOU54F: passing run:- https://github.com/YOU54F/pact-php/actions/runs/9128336180

will get the plugin index updated today :)

tienvx: > pact-php always try and grab the latest version of the plugin on the index, or can a user fix it to a particular version?

pact-php doesn't have any control over which version of plugin to download

YOU54F: hmmm, so the `pactffi_using_plugin` function takes two arguments, one is the version.

https://docs.rs/pact_ffi/latest/pact_ffi/plugins/fn.pactffi_using_plugin.html

What are we passing in there? It's nice to have it default to the latest (I think that it what is is doing if you pass in `NULL` to the 2nd arg)

however a user may want to fix, to a version they have installed on their local machine (to avoid trying to pull the latest) I'll check what happens if we have an earlier version of the plugin already installed, will it use that or just grab the latest.

From the provider verifier side, it looks at the pact file to determine which plugin to install. From what I've noted, every run will give me the latest plugins on the consumer side, and then the same on the provider side (as its reading from the regenned pacts)

YOU54F (approved):
🚀 Let there be Alpine!

Nice to be first of the client libraries consuming the FFI to support it properly. 

edit: I'll get the plugin index updated shortly (probably in about an hour as have a meeting now)

.github/workflows/build.yml
YOU54F: We can address the plugin version stuff in a different PR at another time rather than complicating this more

tienvx: > What are we passing in there? It's nice to have it default to the latest (I think that it what is is doing if you pass in NULL to the 2nd arg)

pact-php is using `NULL`

> I'll check what happens if we have an earlier version of the plugin already installed, will it use that or just grab the latest.

I remember I saw the behavior that it will use the version that's already installed, doesn't grab the latest.

YOU54F: Thanks buddy.

Index now updated with the latest csv and avro plugins

https://github.com/pact-foundation/pact-plugins/pull/62

PR #618, opened 2024-06-20
ci: Upgrade PHP to 8.3

tienvx:
* I'm not sure why setting up PHP 8.1 on Ubuntu 22.04 taking so long, probably because it need to compile grpc extension from source code. I think the default PHP version changed to 8.3, but the [document](https://github.com/shivammathur/setup-php?tab=readme-ov-file#github-hosted-runners) is not updated.
  * Solution: Upgrade PHP to 8.3
* Look like grpc extension is available for PHP 8.3 on Windows https://pecl.php.net/package/gRPC
  * Change: Upgrade PHP to 8.3 to make tests run faster, since PHP 8.3 is default version on Windows
* There are no default version of PHP installed on macos-14 https://github.com/shivammathur/setup-php?tab=readme-ov-file#github-hosted-runners
  * So upgrade PHP from 8.2 to `8.3` doesn't make the tests run slower

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,46 +44,31 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        php: [8.3]
+        os: [ubuntu-latest, macos-12, macos-14]
         include:
-          - os: ubuntu-latest
-            php: 8.1
-          - os: macos-12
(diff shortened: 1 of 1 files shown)


YOU54F (approved):
gotta go fast :)

PR #508, opened 2024-03-03
feat: Allow to build multiple interactions

tienvx:
* Allow to build multiple interactions
* Convert `InteractionBuilderTest` to unit tests with mocks

This is how end-user use it:
```php
$builder = new InteractionBuilder($config);
$builder
       ->given('a person exists')
       ->uponReceiving('a get request to /hello/{name}')
       ->with($request)
       ->willRespondWith($response, false);
$builder
       ->newInteraction()
       ->given('a person exists')
       ->uponReceiving('a get request to /goodbye/{name}')
       ->with($request)
       ->willRespondWith($response);
```

If not using this feature (nothing change):

```php
$builder = new InteractionBuilder($config);
$builder
       ->given('a person exists')
       ->uponReceiving('a get request to /hello/{name}')
       ->with($request)
       ->willRespondWith($response);
```

PR #620, opened 2024-07-29
docs: document multiple interactions

individual-it:
https://github.com/pact-foundation/pact-php/pull/508 added the support to have multiple independent interactions with one builder, but:
- it was not documented
- the examples in the PR did not work for me (maybe the code got changed since then) :man_shrugging: 

This PR adds documentation for this super useful feature and adds an example


YOU54F (approved):
Thanks @individual-it!

PR #623, opened 2024-08-29
Fix wrong content type in protobuf plugin 0.5

tienvx:
Fix this error if using protobuf plugin version 0.5 (not sure about 0.4)

```
expected a body of 'application/protobuf;message=.library.Person' but the actual content type was 'application/protobuf;message=Person'
```

I guess the `.library.` and `.plugins.` prefixes come from the `package library;` and `package plugins;` in proto files.

Depend on https://github.com/pact-foundation/pact-php/pull/622

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "friendsofphp/php-cs-fixer": "^3.0",
         "php-amqplib/php-amqplib": "^3.0",
         "phpstan/phpstan": "^1.9",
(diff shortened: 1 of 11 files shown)


YOU54F:
yep issue is from 0.5.0 onwards following an update

https://github.com/pactflow/pact-protobuf-plugin/releases/tag/v-0.5.0

YOU54F (approved):

PR #627, opened 2024-08-30
feat: Allow add custom provider headers

tienvx:
Add ability to add custom provider headers to v10. Somehow I didn't notice this feature is available in v9 and FFI at the same time.


YOU54F (approved):
Looks good,

I would have probably used a different example, rather than introducing a new one with some matcher todos. (Adding the form encoded example, may be better in a sep PR)

Update on docs would be nice - https://github.com/pact-foundation/pact-php/blob/master/docs/provider.md maybe pointing to this 

https://docs.pact.io/provider/handling_auth#4-modify-the-request-to-use-real-credentials

### Modify Requests Prior to Verification 

and borrow some text from pact-js as to the rationale

https://github.com/pact-foundation/pact-js/blob/master/docs/provider.md#modify-requests-prior-to-verification-request-filters

I don't believe we have request filters as a concept in pact-php for modification of payloads prior to being sent to the provider. I believe JVM and JS do this by starting up a proxy.

tienvx:
I added the doc and removed the `form-urlencoded` example. But I am not confident adding to any existing example

PR #633, opened 2024-09-22
refactor: Split autoload for each example

tienvx:
Maintain a giant list of autoload items in single file for every examples is not ideal. We need to change this file everytime we add new example.

This PR move autoload items into each example. Now it can load only few required items when needed.


YOU54F:
makes sense to me, would make it easier for a user to lift and shift an example into their own project

tienvx:
* Because PHPUnit doesn't support specify bootstrap script for each testsuite (https://github.com/sebastianbergmann/phpunit/issues/3795), I need to remove those PHPUnit test suites for examples, and adding Composer's scripts instead
* Because of it, I have to modify Github Actions's workflow to run those Composer's scripts instead of filtering PHPUnit's test suites.
* I also try to run all examples at once on Windows. And :magic_wand:  : tests doesn't fail as usual. I think it because of pact plugins cache has been cleared
* Upgrade `macos-12` to `macos-13`, because of timeout. Look like `macos-12` is not supported anymore by `setup-php` https://github.com/shivammathur/setup-php?tab=readme-ov-file#github-hosted-runners

JP-Ellis (approved):

tienvx:
Code is moved to #637 to make this PR smaller

Issue #672, opened 2024-10-01
Random fail (timeout) on provider example (mostly Windows)


tienvx:
```
OK, but there were issues!
Tests: 1, Assertions: 2, Deprecations: 1.
PHPUnit 11.3.6 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.11
Configuration: D:\a\pact-php\pact-php\example\binary\provider\phpunit.xml

Script cd example/binary/provider && phpunit handling the run-example:binary event returned with error code 2
E                                                                   1 / 1 (100%)
err > [Thu Sep 26 03:19:34 2024] PHP 8.3.11 Development Server (http://127.0.0.1:50239/) started
Script composer run-example:binary handling the run-examples event returned with error code 2


Time: 01:00.306, Memory: 8.00 MB

There was 1 error:

1) BinaryConsumer\Tests\PactVerifyTest::testPactVerifyConsumer
Symfony\Component\Process\Exception\ProcessTimedOutException: The process "php -S 127.0.0.1:50239 -t "D:\a\pact-php\pact-php\example\binary\provider\tests/../public/"" exceeded the timeout of 60 seconds.

D:\a\pact-php\pact-php\vendor\symfony\process\Process.php:1181
D:\a\pact-php\pact-php\vendor\symfony\process\Process.php:491
D:\a\pact-php\pact-php\tests\PhpPact\Helper\AbstractProcess.php:28
D:\a\pact-php\pact-php\example\binary\provider\tests\PactVerifyTest.php:17
```

Issue #641, opened 2024-09-24
Compatibility Suite: Random fail on 'V3 era Generators applied to HTTP parts' feature


tienvx:
```
Feature: V3 era Generators applied to HTTP parts

  Scenario: Supports using a generator with the request path             # compatibility-suite/pact-compatibility-suite/features/V3/http_generators.feature:3
    Given a request configured with the following generators:            # PhpPactTest\CompatibilitySuite\Context\V3\RequestGeneratorsContext::aRequestConfiguredWithTheFollowingGenerators()
      | generators                                                                 |
      | JSON: { "path": { "type": "ProviderState", "expression": "/path/${id}" } } |
    And the generator test mode is set as "Provider"                     # PhpPactTest\CompatibilitySuite\Context\V3\RequestGeneratorsContext::theGeneratorTestModeIsSetAs()

Verifying a pact between c and p

  Interaction 1 (0s loading, 1s 113ms verification)
      Request Failed - One or more of the setup state change handlers has failed


Failures:

1) Verifying a pact between c and p - Interaction 1 - One or more of the setup state change handlers has failed


There were 1 pact failures

    When the request is prepared for use with a "providerState" context: # PhpPactTest\CompatibilitySuite\Context\V3\RequestGeneratorsContext::theRequestIsPreparedForUseWithAProviderStateContext()
      | { "id": 1000 } |
    Then the request "path" will be set as "/path/1000"                  # PhpPactTest\CompatibilitySuite\Context\V3\RequestGeneratorsContext::theRequestWillBeSetAs()
      Warning: file_get_contents(compatibility-suite/tests/Constant/../../public/generators/path.txt): Failed to open stream: No such file or directory in compatibility-suite/tests/Service/GeneratorServer.php line 42

```

Issue #549, opened 2024-04-12
Compatibility Suite: Random fail on 'Supports a regex generator' scenario


tienvx:
```
Feature: V3 era Generators

...


  Scenario: Supports a regex generator                                                     # compatibility-suite/pact-compatibility-suite/features/V3/generators.feature:31
    Given a request configured with the following generators:                              # PhpPactTest\CompatibilitySuite\Context\V3\RequestGeneratorsContext::aRequestConfiguredWithTheFollowingGenerators()
      | body             | generators                 |
      | file: basic.json | randomregex-generator.json |

Verifying a pact between c and p

  Interaction 1 (0s loading, 5ms verification)
    returns a response which
      has status code 200 (FAILED)
      has a matching body (OK)


Failures:

1) Verifying a pact between c and p - Interaction 1
    1.1) has status code 200
           expected 200 but was 500

There were 1 pact failures

    When the request is prepared for use                                                   # PhpPactTest\CompatibilitySuite\Context\V3\RequestGeneratorsContext::theRequestIsPreparedForUse()
    Then the body value for "$.one" will have been replaced with a "string from the regex" # PhpPactTest\CompatibilitySuite\Context\V3\BodyGeneratorsContext::theBodyValueForWillHaveBeenReplacedWithA()
      string does not contain a valid JSON object. (JsonPath\InvalidJsonException)

```

https://github.com/tienvx/pact-php/actions/runs/11158029525/job/31013436962

tienvx:
This issue is not resolved by the PR above

Issue #260, opened 2022-05-28
Newer syntax support


Lewiscowles1986:
Assuming https://github.com/pact-foundation/pact-php/issues/258 can be agreed upon.
As seen in https://github.com/pact-foundation/pact-php/pull/210 php 7.4, 8& 8.1 can provide new syntax enhancements.

Technically 7.4 will be EOL very soon.

This is a placeholder discussion for discussing the use of more modern PHP syntax.

tienvx:
I tried to use many PHP 8.1's features as possible listed at https://stitcher.io/blog/new-in-php-81

I think we can close this issue

PR #673, opened 2024-10-01
refactor: Replace Slim Framework by Framework X

tienvx:
Try to fix these issues:
* #672
* #641
* #549

[Slim Framework](https://www.slimframework.com/) + [PHP Built-in web server](https://www.php.net/manual/en/features.commandline.webserver.php) -> slow
> The web server runs only one single-threaded process, so PHP applications will stall if a request is blocked. 

[Framework X](https://framework-x.org/) + [ReactPHP](https://reactphp.org/) -> faster 2x -> 4x
> Supports async and non-blocking execution for maximum performance

Another option is [AMPHP](https://amphp.org/http-server-router), but it's a bit slower, and the syntax is not  similar to Slim Framework than Framework X

### DIFF
--- a/compatibility-suite/public/generators/index.php
+++ b/compatibility-suite/public/generators/index.php
@@ -1,33 +1,30 @@
 <?php
 
-use Psr\Http\Message\ResponseInterface as Response;
-use Psr\Http\Message\ServerRequestInterface as Request;
-use Slim\Factory\AppFactory;
+use React\Http\Message\Response;
+use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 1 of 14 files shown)


tienvx:
Tested with:
* Re-run compatibility suite v3 10 times
* Re-run examples on Windows 10 times
* Re-run examples on MacOS 13 5 times
* Re-run examples on MacOS 14 5 times

JP-Ellis (approved):

tienvx:
Indirectly related to #260

PR #707, opened 2024-12-15
deps: Tell renovate to ignore docker-compose.yml

tienvx:
`docker-compose.yml` is not used anywhere. It's just an example file. We can tell Renovate to ignore that file, prevent it from keep pinning dependencies on that file.


YOU54F (approved):

JP-Ellis:
If the docker compose isn't used anywhere, should we just remove it?

Or if it is used as an example, shouldn't we keep updating it (albeit, perhaps without pinning dependencies)?

PR #714, opened 2025-01-08
chore(ci): prevent duplicate ci triggers

JP-Ellis:
Triggering on all 'push' and 'pull_request' events will result in duplicated CI runs in PRs (due to both conditions being met).

Instead, trigger only on pushes to `master`, or updates to PRs targetting `master`.

Also removed cron job as all updates to `master` are checked anyway.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,10 +2,11 @@ name: Code Analysis & Test
 
 on:
   push:
+    branches:
+      - master
   pull_request:
-  # Once on the first of the month at 06:00 UTC
-  schedule:
-    - cron: 0 6 1 * *
+    branches:
+      - master
 
 concurrency:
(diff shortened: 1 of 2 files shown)


tienvx (approved):
Thank Josh.

I tried to solve this problem before but I don't remember why I stop doing that.

.github/workflows/build.yml
```diff
@@ -2,10 +2,11 @@ name: Code Analysis & Test
 
 on:
   push:
+    branches:
+      - master
   pull_request:
-  # Once on the first of the month at 06:00 UTC
-  schedule:
-    - cron: 0 6 1 * *
```
tienvx, line 8: I think this code is for early detecting if any dependencies breaking the CI. I think we can remove it

PR #732, opened 2025-05-31
test: Test repeated matching for protobuf

tienvx:


tienvx:
Found a related issue, created here https://github.com/pactflow/pact-protobuf-plugin/issues/179

PR #735, opened 2025-06-01
test: Test provider state generator for protobuf

tienvx:


tienvx:
I wonder why provider doesn't receive the value (id) from provider state. Probably it's another bug?

tienvx:
Created an issue at https://github.com/pactflow/pact-protobuf-plugin/issues/191

PR #751, opened 2025-10-05
chore(ci): Replace macos 13 runner

tienvx:


YOU54F:
I _think_ the macos-15-intel is a paid runner?

Can we work around it with x64 emulation (using rosetta) in an arm64 runner?

